Issue #42, opened 2021-10-01
Mirror/Proxy APM ServerInformation Endpoint


astorm:
APM Server has [a server information endpoint](https://www.elastic.co/guide/en/apm/server/current/server-info.html).

If called with an secret token to an API Key, this endpoint will return information about the server

```
# curl -H 'Authorization: Bearer ${secretToken}' http://apm-server.example.com:8200
% curl -H 'Authorization: ApiKey ${apiKey}' http://apm-server.example.com:8200
{
  "build_date": "2021-03-18T05:15:57Z",
  "build_sha": "ef735c0cc1e5aa0d4fe4a363804390776e7775fe",
  "version": "7.12.0"
}
```
If called without a token or api key, the endpoint will return a status code of 200.

```
# note the -i option for returning headers as well as the request body (the later of which is empty)
% curl -i http://apm-server.example.com:8200
HTTP/2 200 
date: Fri, 01 Oct 2021 19:28:52 GMT
x-cloud-request-id: J...w
x-content-type-options: nosniff
x-found-handling-cluster: 0...8
x-found-handling-instance: instance-0000000012
content-length: 0
```

Some agents use this as a health endpoint or version gating.  

We should considering exposing the same endpoint in the extension and proxying this endpoint to the final APM Server destination.

PR #52, opened 2021-10-13
feat: Proxy APM Server Info Request

astorm:
Implements: #42

This PR implements a new `/` endpoint for the extension's main HTTP server.  Calling this endpoint will forward/proxy a request to APM Server's `/` endpoint.  The intended use is to allow agents that use this URL as a health check/version fetcher will be able to do so.

In addition to the above functionality, this PR also

- Changes `config.apmServerEndpoint` to `config.apmServerUrl` and moves the logic for creating the full `localhost:8200/intake/v2/events` URL out of the configuration/env parsing code
- When parsing `ELASTIC_APM_LAMBDA_APM_SERVER`, we now ensure it always ends in a `/`
- Implements a test case that uses a Mocked APM Server to confirm that requests are forwarded/proxied correctly


apm-lambda-extension/extension/http_server_test.go
```diff
@@ -137,3 +143,116 @@ func Test_getDecompressedBytesFromRequestEmptyBody(t *testing.T) {
 		t.Fail()
 	}
 }
+
```
estolfo, line 146: Is there a reason you didn't use the [httptest](https://pkg.go.dev/net/http/httptest) package here?
It might be a lot easier to write this test using it. 

Here is an example in the apm server repo: https://github.com/elastic/apm-server/blob/master/sourcemap/fleet_store_test.go

You could create a mock server using httptest for the apm server and write a handler that asserts directly on the request values. I know it would mean writing this all over again but I think it's the standard way to test http servers.

Let me know if I'm missing something and you'd like to stick with this approach.

apm-lambda-extension/extension/http_server_test.go
```diff
@@ -137,3 +143,116 @@ func Test_getDecompressedBytesFromRequestEmptyBody(t *testing.T) {
...
+		apmServerSecretToken:       "foo",
+		apmServerApiKey:            "bar",
+		dataReceiverServerPort:     ":" + extensionPort,
+		dataReceiverTimeoutSeconds: 15,
+	})
+
+	// sleep seem neccesary in our test env
+	// to give the above go-routines-servers
+	// time to complete their startup
+	time.Sleep(1 * time.Second)
```
estolfo, line 237: I've seen in the apm server repo that they use `WaitGroup`s for this. I think it would be a preferable approach over `Sleep`ing.

PR #72, opened 2021-11-08
Preserve compressed agent data and compress agent data that is not already compressed

### LINKED ISSUE #71: Don't decompress agent data
When agents send compressed data, the extension should not decompress it, just to compress it again when sending the data to APM Server.

Buffering compressed data is also lighter on memory.

The extension needs to remember the `Content-Encoding` header the agent has sent and use the same when forwarding the data to APM Server.

See also https://github.com/elastic/apm/pull/518#issuecomment-948538762

estolfo:
This PR resolve #71 

The changes include:
- Create a struct for representing the agent data. It has members for the agent data and content encoding of the request
- Change the type of the agent data channel to the `AgentData` struct
- Don't decompress the agent data when it is received by the extension
- When the extension sends the data to the apm server, compress data that is not already compressed and just pass along the content encoding and unchanged data otherwise.
- Remove the `getDecompressedBytesFromRequest` function, as we won't need it anymore
- Remove the tests for the `getDecompressedBytesFromRequest` function


apm-lambda-extension/extension/apm_server.go
```diff
@@ -28,28 +28,37 @@ import (
...
+	var err error
 
-	client := &http.Client{}
-
-	req, err := http.NewRequest("POST", config.apmServerUrl+endpointUri, bytes.NewReader(compressedBytes.Bytes()))
-	if err != nil {
-		return fmt.Errorf("failed to create a new request when posting to APM server: %v", err)
+	if agentData.ContentEncoding == "" {
+		var compressedBytes bytes.Buffer
+		w := gzip.NewWriter(&compressedBytes)
```
estolfo, line 38: @felixbarny Should we use gzip to compress the agent data that is not already compressed when sending to the apm server?

felixbarny: GZip is fine but we should compress with the best speed compression level

apm-lambda-extension/extension/apm_server.go
```diff
@@ -28,28 +28,37 @@ import (
...
+	var req *http.Request
+	var err error
 
-	client := &http.Client{}
-
-	req, err := http.NewRequest("POST", config.apmServerUrl+endpointUri, bytes.NewReader(compressedBytes.Bytes()))
-	if err != nil {
-		return fmt.Errorf("failed to create a new request when posting to APM server: %v", err)
+	if agentData.ContentEncoding == "" {
+		var compressedBytes bytes.Buffer
```
felixbarny, line 37: Use an `io.Pipe()` to avoid the temporary buffer.

apm-lambda-extension/extension/route_handlers.go
```diff
@@ -68,13 +73,18 @@ func handleInfoRequest(handler *serverHandler, w http.ResponseWriter, r *http.Re
 
 // URL: http://server/intake/v2/events
 func handleIntakeV2Events(handler *serverHandler, w http.ResponseWriter, r *http.Request) {
-	bodyBytes, err := getDecompressedBytesFromRequest(r)
-	if nil != err {
-		log.Printf("could not get decompressed bytes from request body: %v", err)
+	if r.Body == nil {
+		log.Println("Could not get bytes from agent request body")
 	} else {
+		rawBytes, _ := ioutil.ReadAll(r.Body)
```
astorm, line 79: It looks like `ioutil.ReadAll` doesn't promise success.  Should we be handling the error here instead of suppressing with an `_`? Or is that not needed?

astorm (commented):
Looks like a great start and the general shape of things looks correct to me.  Two things to address and (I think) we'll be set 

- I've not done it before but Felix's suggestion of using an `io.Pipe()` to avoid the `compressedBytes` buffer variable is sound

- I have a question about error handling in the new code that reads the agent request body that might lead to a little extra work (or an explanation)

apm-lambda-extension/extension/route_handlers.go
estolfo: I'll ask what the apm server does, as we should probably handle the error similarly to how they would.

apm-lambda-extension/extension/apm_server.go
estolfo: done

estolfo:
@felixbarny If the extension tries to compress non-compressed agent data and fails, what should the extension do? Send the data non-compressed?

apm-lambda-extension/extension/route_handlers.go
estolfo: https://github.com/elastic/apm-server/blob/master/beater/api/intake/handler.go#L72-L100

If the apm server can't read the request, it doesn't log the error but rather writes the error back to the client. I'm not sure we would want to write errors back to the agent, as the agent can't do anything more with the information than the extension can, and we don't want to hold up the lambda function. It seems to me like the best way to handle this would be to log the error and move on. @astorm what do you think? Maybe @felixbarny has some thoughts as well.

felixbarny:
The compression happens while sending data to APM Server. That means if an error happens, some data might have already been sent to the Server. I think the primary source of errors would be network issues rather than issues related to compressing the data. In these cases, I think it's fine to log the error and discard the data.

apm-lambda-extension/extension/route_handlers.go
felixbarny: SGTM. If we find use cases that require sending a response to the agent, we can always add it later.

apm-lambda-extension/extension/apm_server.go
```diff
@@ -21,35 +21,53 @@ import (
...
 
-	req, err := http.NewRequest("POST", config.apmServerUrl+endpointUri, bytes.NewReader(compressedBytes.Bytes()))
-	if err != nil {
-		return fmt.Errorf("failed to create a new request when posting to APM server: %v", err)
+		go func() {
+			_, err = io.Copy(gw, bytes.NewReader(agentData.Data))
+			gw.Close()
+			pw.Close()
+			if err != nil {
+				log.Printf("Failed to compress data: %v", err)
```
felixbarny, line 46: [minor suggestion]
IIUC, this function will only be executed once the request has been established and is then streamed (via chunked-encoding) to the server.
If that's correct, the message may be a bit misleading as the more common source of failure is that there's a network issue when streaming the data to APM Server.

```suggestion
				log.Printf("Failed to send compressed data: %v", err)
```

felixbarny (approved):

apm-lambda-extension/extension/route_handlers.go
astorm: @estolfo logging the error and moving on seem sufficient to me as well.

astorm (commented):
Happy to approve once we have logging for the error case on 
```
rawBytes, _ := ioutil.ReadAll(r.Body)
```

apm-lambda-extension/extension/apm_server.go
estolfo: We are actually not streaming data yet, but rather sending in batches. At this point in the code, the actual failure is that the data could not be compressed in a go routine. If compression failed in that go routine, we will also get an error that is logged on line 52 (`failed to create a new request when posting to APM server`) and the function will return. So I think the error message here is accurate and should stay as-is.

felixbarny: I think this streams the batches. See also https://medium.com/stupid-gopher-tricks/streaming-data-in-go-without-buffering-3285ddd2a1e5#2342

estolfo: Yes, you're right that the actual send could be chunked by the go internal transport code, I thought you were referring to the way that we send the data as it's received from the agent. But at this point in the code, I'm calling `io.Copy`, which ends up calling [Write](https://pkg.go.dev/compress/gzip#Writer.Write) on the `gzipWriter` before it gets to the network code. So any compression failures would be returned by `io.Copy`. Network errors would be returned by [resp, err := client.Do(req)](https://github.com/elastic/apm-aws-lambda/pull/72/files#diff-6c889602ebc4f41bed697c18ef699798a40df79b104d827330a8bc38be305324R71).

But that actually brings up another point: the error returned from `client.Do(req)` actually wraps the compression error, which means we don't really need to do anything with the error returned from `io.Copy`. In other words, we could removing the logging of the error and just rely on the error returned in [this line](https://github.com/elastic/apm-aws-lambda/pull/72/files#diff-6c889602ebc4f41bed697c18ef699798a40df79b104d827330a8bc38be305324R73)

felixbarny: > I'm calling io.Copy, which ends up calling Write on the gzipWriter before it gets to the network code.

What I'm understanding from the blog that I linked is that `io.Copy` blocks until the `PipeReader` is being read from. This happens when `client.Do(req)` is invoked and the data from the `PipeReader` is written into the HTTP request body via chunked encoding. It's because `io.Copy` blocks that we have to execute it in a concurrently running goroutine. If we didn't do that, we'd never get to the `client.Do(req)` part.

See also the docs for [`io.Pipe`](https://pkg.go.dev/io#Pipe)
> each Write to the PipeWriter blocks until it has satisfied one or more Reads from the PipeReader that fully consume the written data. The data is copied directly from the Write to the corresponding Read (or Reads); there is no internal buffering.

PR #73, opened 2021-11-09
fix: use "amd64" string for intel-compat builds (amazon uses x86_64)

astorm:
This fixes a bug in the installer's chip architecture detector. 

Amazon uses the following abbreviations for ARM/Intel: `arm64`, `x86_64`

Go's build tools use the following abbreviations for ARM/Intel: `arm64`, `amd64`

This PR make sure that we're setting the correct value when building the extension with the installer.


estolfo (approved):
🙃  thanks for figuring this out

Issue #62, opened 2021-10-21
Use explicit request from agent to signal completion of lambda function


estolfo:
Instead of relying on the logs API to send a notification that the function has completed, use a specific request from the agent to the extension.

Whether the request is one to the intake API with a particular query param or if it's a specific endpoint, is TBD.

PR #78, opened 2021-11-16
Refactor route handling and server

estolfo:
While working on #62, I found that it was awkward to find a way to pass the channel signaling that the function completed to the router handler. In defining the handlers as they are in this PR, we can easily pass the channel to the handler function, without having to keep a reference to it somewhere.

### DIFF
--- a/apm-lambda-extension/extension/http_server.go
+++ b/apm-lambda-extension/extension/http_server.go
@@ -19,24 +19,26 @@ package extension
 
 import (
 	"log"
+	"net"
 	"net/http"
 )
 
 var agentDataServer *http.Server
 
-func StartHttpServer(agentDataChan chan AgentData, config *extensionConfig) {
+func StartHttpServer(agentDataChan chan AgentData, config *extensionConfig) (err error) {
 	mux := http.NewServeMux()
 	mux.HandleFunc("/", handleInfoRequest(config.apmServerUrl))
 	mux.HandleFunc("/intake/v2/events", handleIntakeV2Events(agentDataChan))
 	agentDataServer = &http.Server{Addr: config.dataReceiverServerPort, Handler: mux}
 
+	ln, err := net.Listen("tcp", agentDataServer.Addr)
+	if err != nil {
+		return
+	}
+
 	go func() {
+		agentDataServer.Serve(ln)
 		log.Printf("Extension listening for apm data on %s", agentDataServer.Addr)
-		err := agentDataServer.ListenAndServe()
-		if err != http.ErrServerClosed {
-			log.Printf("Unexpected stop on Extension Server: %v", err)
-		} else {
-			log.Printf("Extension Server closed %v", err)
-		}
 	}()
+	return nil
 }
--- a/apm-lambda-extension/extension/http_server.go
+++ b/apm-lambda-extension/extension/http_server.go
@@ -37,8 +37,8 @@ func StartHttpServer(agentDataChan chan AgentData, config *extensionConfig) (err
 	}
 
 	go func() {
-		agentDataServer.Serve(ln)
 		log.Printf("Extension listening for apm data on %s", agentDataServer.Addr)
+		agentDataServer.Serve(ln)
 	}()
 	return nil
 }
--- a/apm-lambda-extension/extension/http_server.go
+++ b/apm-lambda-extension/extension/http_server.go
@@ -18,50 +18,25 @@
 package extension
 
 import (
-	"net"
+	"log"
 	"net/http"
-	"time"
 )
 
-type serverHandler struct {
-	data   chan AgentData
-	config *extensionConfig
-}
+var extensionServer *http.Server
 
-func (handler *serverHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	if r.URL.Path == "/intake/v2/events" {
-		handleIntakeV2Events(handler, w, r)
-		return
-	}
+func StartHttpServer(agentDataChan chan AgentData, config *extensionConfig) {
+	mux := http.NewServeMux()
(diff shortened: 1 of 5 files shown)


apm-lambda-extension/extension/http_server.go
```diff
@@ -18,50 +18,27 @@
...
-		Handler:        &handler,
-		ReadTimeout:    timeout,
-		WriteTimeout:   timeout,
-		MaxHeaderBytes: 1 << 20,
-	}
+func StartHttpServer(agentDataChan chan AgentData, config *extensionConfig) (err error) {
+	mux := http.NewServeMux()
+	mux.HandleFunc("/", handleInfoRequest(config.apmServerUrl))
+	mux.HandleFunc("/intake/v2/events", handleIntakeV2Events(agentDataChan))
+	agentDataServer = &http.Server{Addr: config.dataReceiverServerPort, Handler: mux}
```
astorm, line 32: Question: I noticed we've lost the ability to set/configure the timeouts and `MaxHeaderBytes` here.  Do you know what the defaults are and whether they're appropriate for our use case?

apm-lambda-extension/extension/process_events.go
```diff
@@ -26,6 +26,7 @@ import (
 func ProcessShutdown() {
 	log.Println("Received SHUTDOWN event")
 	log.Println("Exiting")
+	agentDataServer.Close()
```
astorm, line 29: Question: No objections to it but was/is adding this `.Close()` call here just house keeping or did the lack of this have a negative side-effect on the extension?

apm-lambda-extension/extension/http_server_test.go
```diff
@@ -46,15 +47,17 @@ func TestInfoProxy(t *testing.T) {
 		apmServerUrl:               apmServer.URL,
 		apmServerSecretToken:       "foo",
 		apmServerApiKey:            "bar",
-		dataReceiverServerPort:     "127.0.0.1:1234",
+		dataReceiverServerPort:     ":1234",
```
astorm, line 50: Not mandatory, but since we're changing the semantics of this variable/property to include _only_ the port instead of the server and the port, do we want to change its name to `dataReceiverPort`?

astorm (approved):
Looks like a welcome change -- have a ~routing system~ HTTP request multiplexer in place seems like a standard practice and should let us expand the capabilities of the extension further if needed.  

No objections to landing this as but so _approving_, but I do have some small suggestion/questions below.

apm-lambda-extension/extension/process_events.go
estolfo: This was just some house-keeping. I thought we should properly close the server when we know the lambda env is shutting down.

apm-lambda-extension/extension/http_server.go
estolfo: I didn't mean to remove this configuration, thanks for catching that.
In looking at the `http.Server` documentation, it looks like `ReadTimeout` and friends will be set to their zero value, which is 0, if not set otherwise. The documentation says that a 0 value for `ReadTimeout` and `WriteTimeout` and means there will be no timeout.

I'd like to merge this PR so I can move forward with the explicit agent endpoint but I'll open a second PR to add this configurations back in so we can discuss there what config options are appropriate and if we want to stick with the defaults.

apm-lambda-extension/extension/http_server_test.go
estolfo: I think that makes the most sense. I don't know if users even have the option of starting a server on another host other than localhost from within a lambda environment? I'll open another PR to change this so we can have the discussion there.

apm-lambda-extension/extension/route_handlers.go
```diff
@@ -29,68 +29,73 @@ type AgentData struct {
...
-		log.Printf("could not read info request response to APM Server: %v", err)
-		return
-	}
+		resp, err := client.Do(req)
+		if err != nil {
+			log.Printf("error forwarding info request (`/`) to APM Server: %v", err)
+			return
+		}
+		defer resp.Body.Close()
+		body, err := ioutil.ReadAll(resp.Body)
```
stuartnelson3, line 55: The `body` shouldn't be massive in size, but a good habit is to operate on the `resp.Body` as an `io.Reader`, and use something like `io.Copy` to keep memory allocation down.

apm-lambda-extension/extension/route_handlers.go
```diff
@@ -29,68 +29,73 @@ type AgentData struct {
...
+	return func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusAccepted)
+		w.Write([]byte("ok"))
 
-	rawBytes, err := ioutil.ReadAll(r.Body)
-	if err != nil {
-		log.Println("Could not read bytes from agent request body")
-		return
-	}
+		if r.Body == nil {
```
stuartnelson3, line 83: did you encounter the body being nil? I took a look at the docs because I was checking to see if you need to close the request body, and it says:

```
	// For server requests, the Request Body is always non-nil
	// but will return EOF immediately when no body is present.
	// The Server will close the request body. The ServeHTTP
	// Handler does not need to.
```

so I'm not sure if this nil check is required. I suppose it doesn't hurt, though.

stuartnelson3 (commented):
couple comments but that's all, looks good!

apm-lambda-extension/extension/route_handlers.go
estolfo: thanks, I'll change that 👍🏼

apm-lambda-extension/extension/route_handlers.go
estolfo: If the docs say it will never be `nil`, it doesn't seem necessary. I'll remove it :)

PR #79, opened 2021-11-17
Use more straightforward compression approach and buffer pool

estolfo:
Instead of using `io.Copy`, use a more straightforward compression approach. Also use a buffer pool to avoid creating a new buffer each time a request is sent.

### DIFF
--- a/apm-lambda-extension/extension/apm_server.go
+++ b/apm-lambda-extension/extension/apm_server.go
@@ -35,54 +35,43 @@ var bufferPool = sync.Pool{New: func() interface{} {
 //       connection open across invocations?
 func PostToApmServer(agentData AgentData, config *extensionConfig) error {
 	endpointURI := "intake/v2/events"
-	var request *http.Request
+	encoding := agentData.ContentEncoding
+	buf := bufferPool.Get().(*bytes.Buffer)
+	defer func() {
+		buf.Reset()
+		bufferPool.Put(buf)
+	}()
 
 	if agentData.ContentEncoding == "" {
-		buf := bufferPool.Get().(*bytes.Buffer)
-		defer func() {
-			buf.Reset()
-			bufferPool.Put(buf)
-		}()
-
+		encoding = "gzip"
 		gw, err := gzip.NewWriterLevel(buf, gzip.BestSpeed)
 		if err != nil {
 			return err
 		}
-
 		if _, err := gw.Write(agentData.Data); err != nil {
 			log.Printf("Failed to compress data: %v", err)
 		}
 		if err := gw.Close(); err != nil {
 			log.Printf("Failed write compressed data to buffer: %v", err)
 		}
-
-		req, err := http.NewRequest("POST", config.apmServerUrl+endpointURI, buf)
-		if err != nil {
-			return fmt.Errorf("failed to create a new request when posting to APM server: %v", err)
-		}
-		req.Header.Add("Content-Encoding", "gzip")
-		request = req
 	} else {
-		buf := bufferPool.Get().(*bytes.Buffer)
 		buf.Write(agentData.Data)
-		req, err := http.NewRequest("POST", config.apmServerUrl+endpointURI, buf)
-		if err != nil {
-			return fmt.Errorf("failed to create a new request when posting to APM server: %v", err)
-		}
-		req.Header.Add("Content-Encoding", agentData.ContentEncoding)
-		request = req
 	}
 
-	request.Header.Add("Content-Type", "application/x-ndjson")
+	req, err := http.NewRequest("POST", config.apmServerUrl+endpointURI, buf)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


felixbarny:
Is there a specific issue with `io.Pipe`? I think the advantage of it is that there are no buffers required at all.

estolfo:
Yes, there are no synchronization guarantees and the client could start issuing the request before the gzip writer had finished writing everything. If there's an error, the body would be incomplete. Using a buffer ensures that the data is properly gzipped before it is sent.

apm-lambda-extension/extension/apm_server.go
```diff
@@ -21,54 +21,68 @@ import (
...
+		request.Header.Add("Authorization", "ApiKey "+config.apmServerApiKey)
 	} else if config.apmServerSecretToken != "" {
-		req.Header.Add("Authorization", "Bearer "+config.apmServerSecretToken)
+		request.Header.Add("Authorization", "Bearer "+config.apmServerSecretToken)
 	}
 
 	client := &http.Client{}
-	resp, err := client.Do(req)
+
+	resp, err := client.Do(request)
```
marclop, line 85: I think we can remove `request` and remove the `else` branch:

```go
	endpointURI := "intake/v2/events"
	encoding := agentData.ContentEncoding
	buf := bufferPool.Get().(*bytes.Buffer)
	defer func() {
		buf.Reset()
		bufferPool.Put(buf)
	}()

	if agentData.ContentEncoding == "" {
		encoding = "gzip"
		gw, err := gzip.NewWriterLevel(buf, gzip.BestSpeed)
		if err != nil {
			return err
		}
		if _, err := gw.Write(agentData.Data); err != nil {
			log.Printf("Failed to compress data: %v", err)
		}
		if err := gw.Close(); err != nil {
			log.Printf("Failed write compressed data to buffer: %v", err)
		}
	}

	req, err := http.NewRequest("POST", config.apmServerUrl+endpointURI, buf)
	if err != nil {
		return fmt.Errorf("failed to create a new request when posting to APM server: %v", err)
	}
	req.Header.Add("Content-Encoding", encoding)
	req.Header.Add("Content-Type", "application/x-ndjson")
	if config.apmServerApiKey != "" {
		req.Header.Add("Authorization", "ApiKey "+config.apmServerApiKey)
	} else if config.apmServerSecretToken != "" {
		req.Header.Add("Authorization", "Bearer "+config.apmServerSecretToken)
	}

	client := &http.Client{}
	resp, err := client.Do(req)
```

marclop (commented):
+1 on removing `io.Pipe()` too, I think that going forward we could also have some micro benchmarks to get a good understanding of the impact.

apm-lambda-extension/extension/apm_server.go
estolfo: We would still need to write the agent data bytes to the buffer if the content encoding is "". So we can keep the else but it would simply be:
```go
else {
    buf.Write(agentData.Data)
}
```

marclop (approved):

felixbarny:
> the client could start issuing the request before the gzip writer had finished writing everything. If there's an error, the body would be incomplete.

I don't see an issue with that. In fact, I think that's preferable. It avoids unnecessarily copying from buffer to buffer. The Elastic APM Java agent is also directly serializing the events to the HTTP request body to avoid having to copy and allocate buffers. I have never seen a situation where compression would fail.

marclop:
I modified the PostToAPM function to accept an `*http.Client` and it doesn't cause an allocation every time it is called and [did some benchmarking on my machine](https://gist.github.com/marclop/6fb7021e5ccb5b1350024655298479e0) and the pipes seem to be slower than the pooled `bytes.Buffer` approach most likely because the `io.Pipe()` call causes extra allocations.

Overall it seems to be around 10% faster than the `io.Pipe` approach and it causes less allocations, which is also desirable.

```
$ benchstat old.txt new.txt
name         old time/op    new time/op    delta
PostToAPM-8     151µs ±11%     135µs ± 8%  -10.88%  (p=0.032 n=5+5)

name         old alloc/op   new alloc/op   delta
PostToAPM-8    1.25MB ± 0%    1.22MB ± 0%   -2.46%  (p=0.008 n=5+5)

name         old allocs/op  new allocs/op  delta
PostToAPM-8       112 ± 1%       104 ± 1%   -7.33%  (p=0.008 n=5+5)
```

astorm (approved):
+1 for avoid concurrency complications where we can.  I ran this branch through the usual quick manual smoke test and did not see any regressions -- data continued to flow into my elastic cloud instance. 

No objections to this change going in, but also no strong opinions on which direction we should go (with a slight edit to note that @marclop's quick benchmarking is a an additional `+1` for a move away from the `io.Pipe` approach.

felixbarny (approved):
Thanks for the benchmarks, Marc! I have no objections now.
> I modified the PostToAPM function to accept an *http.Client

Could you add that commit to this PR?

Issue #61, opened 2021-10-20
Rename or Reconsider ELASTIC_APM_EXTENSION_TIMEOUT_SECONDS env Variable


astorm:
The `ELASTIC_APM_EXTENSION_TIMEOUT_SECONDS` env variable allows users to set timeout values 

https://github.com/elastic/apm-aws-lambda/blob/63afa0a37193f0e97f3bed185138bf492eb7f6a2/apm-lambda-extension/extension/http_server.go#L55-L60

We should either

1. Rename this variable to more clear express it's intent, @felixbarny suggested `ELASTIC_APM_EXTENSION_TIMEOUT_SECONDS`

2. Reconsider whether it's needed

This variable was created early in development, and the default value picked was "something a little bit longer than the max time an agent may keep a connection open.  Depending on how consistent the agents are in the env naming, we may be able to look at the value set for `ELASTIC_APM_API_REQUEST_TIME` (or use the default of 10s when it's not set) and just add an additional second to ensure the HTTP Server's timeout is a little bit longer than the max agent requests time.

astorm:
In general I'm right there with you with keeping configuration to a minimum, but here's my general thinking on why we might want the timeouts configurable

1. I'm uneasy with the default timeout of 0 -- I'm concerned about the the server staying open indefinitely and keeping the execution environment unfrozen/live, running up bill time 

2. If a timeout is necessary, allowing users to influence this value also seems necessary.  My thinking is if a user needs to extend this timeout for some unseen reason it doesn't seem reasonable to expect them to wait for us to do a release cycle. 

3. This [isn't the first time we've discussed](https://github.com/elastic/apm-aws-lambda/issues/61) this variable, so per #61 we might also consider using `ELASTIC_APM_API_REQUEST_TIME` "plus a little more" as a value.  This reduces the number of variables a user needs to configure, but also gives them indirect control over the timeout.

PR #81, opened 2021-11-19
Add server configurations back in

estolfo:
I unintentionally removed the server configurations in a previous PR and I'm opening this PR so we can discuss whether we want these settings to be configurable.

Questions:
- [ ] `ReadTimeout` and `WriteTimeout` are set to their zero value (0) if not set otherwise, which means there will be no timeout. Do we want users to be able to set these timeouts?
- [ ] Do we want users to be able to set a different `ReadTimeout` and `WriteTimeout`? Currently, they are set to the same user-configured value.
- [ ] `DefaultMaxHeaderBytes` is what is used for `MaxHeaderBytes` if it's not set otherwise. [Its value is 1 << 20](https://cs.opensource.google/go/go/+/refs/tags/go1.17.3:src/net/http/server.go;l=855), so I don't see the point of hardcoding it ourselves to that value.

I personally think we should provide as few configuration options as possible to the user and only options in later if we need them.

### DIFF
--- a/apm-lambda-extension/extension/http_server.go
+++ b/apm-lambda-extension/extension/http_server.go
@@ -21,6 +21,7 @@ import (
 	"log"
 	"net"
 	"net/http"
+	"time"
 )
 
 var agentDataServer *http.Server
@@ -29,7 +30,14 @@ func StartHttpServer(agentDataChan chan AgentData, config *extensionConfig) (err
 	mux := http.NewServeMux()
 	mux.HandleFunc("/", handleInfoRequest(config.apmServerUrl))
 	mux.HandleFunc("/intake/v2/events", handleIntakeV2Events(agentDataChan))
-	agentDataServer = &http.Server{Addr: config.dataReceiverServerPort, Handler: mux}
+	timeout := time.Duration(config.dataReceiverTimeoutSeconds) * time.Second
+	agentDataServer = &http.Server{
+		Addr:           config.dataReceiverServerPort,
+		Handler:        mux,
+		ReadTimeout:    timeout,
+		WriteTimeout:   timeout,
+		MaxHeaderBytes: 1 << 20,
+	}
 
 	ln, err := net.Listen("tcp", agentDataServer.Addr)
 	if err != nil {


astorm:
In general I'm right there with you with keeping configuration to a minimum, but here's my general thinking on why we might want the timeouts configurable

1. I'm uneasy with the default timeout of 0 -- I'm concerned about the the server staying open indefinitely and keeping the execution environment unfrozen/live, running up bill time 

2. If a timeout is necessary, allowing users to influence this value also seems necessary.  My thinking is if a user needs to extend this timeout for some unseen reason it doesn't seem reasonable to expect them to wait for us to do a release cycle. 

3. This [isn't the first time we've discussed](https://github.com/elastic/apm-aws-lambda/issues/61) this variable, so per #61 we might also consider using `ELASTIC_APM_API_REQUEST_TIME` "plus a little more" as a value.  This reduces the number of variables a user needs to configure, but also gives them indirect control over the timeout.

estolfo:
Those are good points, I agree we should keep the config in for the timeouts. We can discuss/change the name via the separate issue, which you've referenced.
What about `MaxHeaderBytes`? Should we remove that explicit configuration and rely on the default value, which is what we are using now anyway?

astorm (commented):
> What about MaxHeaderBytes? Should we remove that explicit configuration and rely on the default value, which is what we are using now anyway?

Hmm, thinking that one through, making `MaxHeaderBytes` user-configurable seems like too much configuration complication at this point.

Re: whether we need an explicit value or not — that's a "six of one, half dozen the other" decision for me, but talking though those trade-offs

- Having an explicit value of 1MB (all hail bitwise arithmetic: `1 << 20`) means our code is slightly more complicated

- Not having an explicit value means we're relying on a default, which might be different in older versions of Go or change in a future version of go

So in my mind having a default means we're more in control of how our program will behave when/if folks compile it themselves, so I'd be in favor of keeping it.

estolfo:
ok @astorm that works for me 👍🏼

astorm (approved):
👍

PR #82, opened 2021-11-19
Use query param to intake API to signal that function completed

### LINKED ISSUE #62: Use explicit request from agent to signal completion of lambda function

estolfo:
This PR adds a change to signal on a channel when it receives an intake request from an agent containing the query param "flushed=true".
We can change the name of the query param, if that's not clear enough.

As a fallback, it also signals if it receives a `runtimeDone` event from the logsapi.

There are now three channels used to signal in main. The `AgentDoneSignal` signals that there was an intake request received with `flushed=true`, the `runtimeDoneSignal` channel signals that a runtimeDone event was received via the Logs API. The `funcDone` channel is closed if a signal is received on `AgentDoneSignal` or `runtimeDoneSignal`.
Closing that channel signals to the go routines posting apm data and processing logs API events that they can return.

Resolves #62

### DIFF
--- a/apm-lambda-extension/main.go
+++ b/apm-lambda-extension/main.go
@@ -116,8 +116,8 @@ func main() {
 			// Make a channel for signaling that the function invocation is complete
 			funcDone := make(chan struct{})
 
-			// Flush any APM data, in case waiting for the FuncDone signal timed out,
-			// the agent data wasn't available yet, and we got to the next event
+			// Flush any APM data, in case waiting for the agentDone or runtimeDone signals
+			// timed out, the agent data wasn't available yet, and we got to the next event
 			extension.FlushAPMData(client, agentDataChannel, config)
 
 			// A shutdown event indicates the execution environment is shutting down.
@@ -170,7 +170,7 @@ func main() {
 				}
 			}()
 
-			// Calculate how long to wait for a FuncDone signal
+			// Calculate how long to wait for a runtimeDoneSignal or AgentDoneSignal signal
 			flushDeadlineMs := event.DeadlineMs - 100
 			durationUntilFlushDeadline := time.Until(time.Unix(flushDeadlineMs/1000, 0))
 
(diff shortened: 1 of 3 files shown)


apm-lambda-extension/main.go
```diff
@@ -60,12 +60,17 @@ func main() {
...
+	// Start http server to receive data from agent
 	extension.StartHttpServer(agentDataChannel, config)
 
+	// Create a client to use for sending data to the apm server
+	client := &http.Client{
+		Transport: http.DefaultTransport.(*http.Transport).Clone(),
+	}
+
 	// Make channel for collecting logs and create a HTTP server to listen for them
 	logsChannel := make(chan logsapi.LogEvent)
```
astorm, line 75: The PR description mentions _removed all the logsapi code and tests, as we won't need them anymore with these changes._ -- is the PR description out of date, or does this code not belong here? (or some third thing)

astorm (changes requested):
Overall this looks simple and solid.  I see we have a new `FuncDone` property on the extension package that we're using as a signal only channel, and when this channel receives a message it indicates the function is done with execution.  

We send a signal to this channel whenever we see a request to `/intake/v2/events` that includes a `flushed=true` query string parameter.

This _should_ work.  However, when I tested this against a real agent without the query param implemented the extension crashed and interfered with a Node.js Lambda function (causing it to return `null`)   

However we coordinate with the agents to get this query param in we should (probably?) ensure that lambda functions with older agents attached don't crash.  (I've added a discussion topic to the next planning meeting)

apm-lambda-extension/main.go
estolfo: Sorry the description is out of date! I'll update it.

estolfo:
Thanks for the review, @astorm. I'll try to reproduce the error and report back.

astorm (approved):
Approving 👍 

I ran a manual smoke test and the fallback code is working -- data's sent, the extension does not crash.  

The query param checking/signaling code looks like it does what it needs to.

PR #83, opened 2021-11-22
Use copy for info response and add intake test

estolfo:
This PR contains these changes:

- Uses copy as per @stuartnelson3's [suggestion](https://github.com/elastic/apm-aws-lambda/pull/78/files#r753004695) to forward the apm server's info response back to the agent
- It adds an intake test and a test that the extension forwards an error from an info request
- Removes a check that the body is nil, as the documentation says it won't ever be nil

### DIFF
--- a/apm-lambda-extension/extension/route_handlers.go
+++ b/apm-lambda-extension/extension/route_handlers.go
@@ -84,11 +84,6 @@ func handleIntakeV2Events(agentDataChan chan AgentData) func(w http.ResponseWrit
 		w.WriteHeader(http.StatusAccepted)
 		w.Write([]byte("ok"))
 
-		if r.Body == nil {
-			log.Println("No body in agent request")
-			return
-		}
-
 		rawBytes, err := ioutil.ReadAll(r.Body)
 		if err != nil {
 			log.Println("Could not read bytes from agent request body")
--- a/apm-lambda-extension/extension/route_handlers.go
+++ b/apm-lambda-extension/extension/route_handlers.go
@@ -18,6 +18,7 @@
 package extension
 
 import (
+	"io"
 	"io/ioutil"
 	"log"
 	"net/http"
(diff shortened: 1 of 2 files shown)


apm-lambda-extension/extension/route_handlers.go
```diff
@@ -46,30 +47,33 @@ func handleInfoRequest(apmServerUrl string) func(w http.ResponseWriter, r *http.
...
 			log.Printf("could not read info request response to APM Server: %v", err)
 			return
 		}
 
-		// send status code
-		w.WriteHeader(resp.StatusCode)
-
 		// send every header received
-		for name, values := range resp.Header {
+		for name, values := range serverResp.Header {
```
stuartnelson3, line 71: Does this work? I had assumed you need to write the headers to the response before writing the body

estolfo: Thanks for pointing that out. I've added a check in the test that the header received from the server is forwarded to the agent.

apm-lambda-extension/extension/route_handlers.go
```diff
@@ -80,12 +84,8 @@ func handleIntakeV2Events(agentDataChan chan AgentData) func(w http.ResponseWrit
 		w.WriteHeader(http.StatusAccepted)
 		w.Write([]byte("ok"))
 
-		if r.Body == nil {
```
astorm, line 83: Presuming the "r.Body will always be `nil`" docs are this line here: https://github.com/golang/go/blob/18934e11ba6ef2b2f21f091ddf4ab6814dcf1959/src/net/http/request.go#L174

astorm (approved):
👍 I don't have an informed opinion on the specific change, but it looks done in a reasonable way, has test coverage, and a quick smoke test vs. a live lambda function indicates that everything is still working as expected.  Approving.

PR #84, opened 2021-11-23
feat: add eslint to cli project

astorm:
Adds eslint to CLI project, adjusts existing code to pass eslint rules.  

I'm not tied to any particular ruleset, (other than a ruleset that is applied by robots ;)), so suggestions for additional entries to `rules:` or even extending something other than `standard` is welcome.

### DIFF
--- a/apm-lambda-extension/cli/.eslintrc.js
+++ b/apm-lambda-extension/cli/.eslintrc.js
@@ -1,0 +1,16 @@
+module.exports = {
+  env: {
+    browser: true,
+    commonjs: true,
+    es2021: true
+  },
+  extends: [
+    'standard'
+  ],
+  parserOptions: {
+    ecmaVersion: 12
+  },
+  rules: {
+    'no-control-regex': 0
+  }
+}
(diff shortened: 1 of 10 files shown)


trentm (approved):
Nice. Works for me.

> I'm not tied to any particular ruleset

If you wanted to consider prettier, rather than `standard` for code formatting you are welcome to copy https://github.com/trentm/node-dashdash/blob/master/.eslintrc.json
I *think* Kibana is using prettier.

But using `standard` is all good.

> (other than a ruleset that is applied by robots ;))

Is that referring to the `'no-control-regex': 0` you have in .eslintrc.js?

My limited understanding is that if one is using stock `standard` that you could reduce the devDependencies to just `"standard": "..."` and the "scripts" to:

```
    "lint": "standard",
    "lint:fix": "standard --fix",
```

as we have in apm-agent-nodejs.git

estolfo (approved):

astorm:
>> (other than a ruleset that is applied by robots ;))
> Is that referring to the 'no-control-regex': 0 you have in .eslintrc.js?

Nope, just referring to a general principle that style guidelines that aren't enforced by a program/robot will always end up drifting.

PR #90, opened 2021-12-07
feat: Profiler Runner for Multiple Scenarios

astorm:
Adds a profiler runner that's capable of testing multiple scenarios.  

Next steps:

- [x] Auto deploy the latest version of our extension layer
- [x] Document runner

Future, Post this PR steps:

- Auto download any needed tools (`lpt-0.1.jar`) and auto configure
- Post results to some elastic search instance
- Auto run `npm install` (or other installs?) for code in `profile/code`?
- Run regularly somehow


estolfo:
Thanks for starting this @astorm. I watched the screencast and looked at the code. I'm not an expert at nodejs so can't comment on the code itself, unfortunately. The only comment I have is, is there a way to make this more general so we could use it for testing? It's essentially doing the same thing I was trying to do with the sam cli: defining different lambda functions and configurations and running the lambda function. The only difference is, the sam cli allows you to run the function locally.
I don't want to broaden the scope for this task too much, so maybe that's not worth it.

PR #93, opened 2022-01-05
Link to built-docs instead of asciidoc file

bmorelli25:
Now that docs are included in the APM Guide (see https://github.com/elastic/apm-server/pull/6984), we should link to the docs instead of the asciidoc file.


astorm (approved):
👍

PR #98, opened 2022-01-12
Allow for different send strategy via environment variable

### LINKED ISSUE #75: Introduce an optional background dispatching strategy
Currently the extension flushes all received data after having received the invocation end signal (from logs API or from the agents directly). Though this unblocks the response of the lambda request itself (response time of the lambda function), it still blocks the lambda function from receiving the *next* request until the extension has flushed all the data. This has a negative effect on the throughput of the function.

For lambda functions that have a steadily frequent load pattern the extension could delay sending the data to the APM server to the *next lambda request* and do the sending in parallel to the processing of that *next request*. This potentially would improve both the lambda function response time and its throughput.

The downside of this strategy is the risk that some APM data might be delayed or even get lost in case that a lambda function has an infrequent load pattern. Therefore, we *cannot* simply apply the above proposed strategy as the single available and the default one.

This issue is about introducing a config option for the dispatching strategy, that would support:
- *flushing strategy*: current dispatching strategy - this should also be default strategy if the config option is not specified explicitly
- *background strategy*: the strategy proposed above

This would allow users to choose between those strategies to optimize for their use case.

Further potential improvements for the future (post GA) are describe in https://github.com/elastic/apm-aws-lambda/issues/7.

estolfo:
Resolve #75

### DIFF
--- a/apm-lambda-extension/extension/process_env.go
+++ b/apm-lambda-extension/extension/process_env.go
@@ -40,6 +40,7 @@ const (
 	// Background send strategy allows the extension to send remaining buffered
 	// agent data on the next function invocation
 	Background SendStrategy = "background"
+
 	// SyncFlush send strategy indicates that the extension will synchronously
 	// flush remaining buffered agent data when it receives a signal that the
 	// function is complete
--- a/apm-lambda-extension/extension/process_env.go
+++ b/apm-lambda-extension/extension/process_env.go
@@ -21,6 +21,7 @@ import (
 	"log"
 	"os"
 	"strconv"
+	"strings"
 )
 
 type extensionConfig struct {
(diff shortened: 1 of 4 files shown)


astorm (commented):
👍 does the thing it says it does and fulfills the request in #75 

I'd say the only thing left is getting our new env variable added to the docs at https://github.com/elastic/apm-aws-lambda/blob/main/docs/aws-lambda-extension.asciidoc#lambda_env and we can call this one good and merge it.

estolfo:
thanks for the reminder about the docs!

estolfo:
@astorm Updated the PR with documentation for the new env variable 👍🏼

PR #106, opened 2022-02-01
Use WaitGroup to ensure go routine finishes sending to APM server

estolfo:
Ensure that the go routine finishes sending to the APM server before the extension considers its work done and polls for the next event.


jlvoiseux (approved):
Also passes the end-to-end tests developed on branch `jlvoiseux/e2e-testing`

Issue #110, opened 2022-02-01
Published ARN Names should Include Release Version and "Elastic"


astorm:
Currently, our build scripts create an extension ARN with a name like

    arn:aws:lambda:us-west-2:123456789123:layer:apm-lambda-extension:80

Amazon automatically embeds the _layer_ version into this ARN (above this is `80`)

We'd like to include the version of the Lambda extension in this layer as well. That is, if we were publishing version `0.0.3` of the extension, we would want to end up with layer name that looked like 

    arn:aws:lambda:us-west-2:123456789123:layer:elastic-apm-lambda-extension-ver-0-0-3:80

The intent of this change would be to tie the name of the ARN with the version of the extension, allowing us to quickly identify which version of the extension's being used without keeping an "ARN version to Extension version" map. It would also bring us in line with what ADOT does with its OpenTelemetry layers ([example](https://aws-otel.github.io/docs/getting-started/lambda/lambda-js))

If this isn't possible we'd like to include the extension version in the table we're publishing of the ARNs so that a specific ARN can be mapped back to a specific extension version.

- [ ] Are we including the extension version in the ARN name, or are we publishing the layer version alongside the ARN in the ASCII doc table
- [ ] Do what we're doing 🤠

trentm:
@astorm 

1. Can we please add "elastic" to the layer name?  Perhaps something like `arn:aws:lambda:<region>:123456789123:layer:elastic-apm-extension-ver-0-0-3:<ver>`.    The "lambda" part of the name could be elided if we want to be shorter, given that "lambda" is required earlier in the full ARN.
2. Do we have separate builds for the separate architectures? If so, do we need the arch in the name as well?

(Happy to make this a separate ticket(s) if you prefer.)

astorm:
@trentm no objections to including elastic in the name and it seems to make sense to do it in this issue -- it should just be a string change (I'll leave @v1v to let us know if I'm missing something obvious)

Also, we do need to publish different ARN for x86 and arm and might already be accounted for in the same -- see the open PR for ARN naming.  https://github.com/elastic/apm-aws-lambda/pull/101

PR #115, opened 2022-02-02
publish: add elastic prefix in the layer name and version as suffix

v1v:
## What

Add elastic prefix in the layer name, version as suffix. Additionally, it adds the arch as suffix too.

### Implementation details

CI is setting the env variable `ELASTIC_LAYER_NAME` dynamically if a new tag release is created.

## Issue

Closes https://github.com/elastic/apm-aws-lambda/issues/110

## Test

image

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -9,7 +9,7 @@ pipeline {
     BASE_DIR = "src/github.com/elastic/${env.REPO}"
     PIPELINE_LOG_LEVEL = 'INFO'
     SUFFIX_ARN_FILE = 'arn-file.md'
-    ELASTIC_LAYER_NAME = 'apm-lambda-extension'
+    ELASTIC_LAYER_NAME = 'elastic-apm-lambda-extension'
   }
   options {
     timeout(time: 1, unit: 'HOURS')
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -179,7 +179,7 @@ pipeline {
 */
 def getVersion() {
   if (env.BRANCH_NAME?.trim() && env.BRANCH_NAME.startsWith('v')) {
-    return env.BRANCH_NAME.replaceAll('v', '-ver').replaceAll('\.', '-')
+    return env.BRANCH_NAME.replaceAll('v', '-ver').replaceAll('\\.', '-')
   }
   return ''
 }
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -9,6 +9,9 @@ pipeline {
     BASE_DIR = "src/github.com/elastic/${env.REPO}"
     PIPELINE_LOG_LEVEL = 'INFO'
     SUFFIX_ARN_FILE = 'arn-file.md'
+    RELEASE_URL_MESSAGE = "(<https://github.com/elastic/${env.REPO}/releases/tag/${env.TAG_NAME}|${env.TAG_NAME}>)"
+    SLACK_CHANNEL = '#apm-serverless'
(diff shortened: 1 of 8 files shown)


apm-lambda-extension/cli/build-and-publish.js
```diff
@@ -38,7 +38,7 @@ function getLastJsonFromShellOutput (output) {
 function buildAndPublish () {
   return new Promise(function (resolve, reject) {
     if (!process.env.ELASTIC_LAYER_NAME) {
-      process.env.ELASTIC_LAYER_NAME = 'apm-lambda-extension'
+      process.env.ELASTIC_LAYER_NAME = 'elastic-apm-lambda-extension'
```
felixbarny, line 41: Maybe removing `lambda` from the name as it's implied from the context and also included in the arn name. See also https://github.com/elastic/apm-aws-lambda/issues/110#issuecomment-1027396444.
```suggestion
      process.env.ELASTIC_LAYER_NAME = 'elastic-apm-extension'
```

.ci/Jenkinsfile
```diff
@@ -30,6 +29,7 @@ pipeline {
         deleteDir()
         gitCheckout(basedir: "${BASE_DIR}")
         stash allowEmpty: true, name: 'source', useDefaultExcludes: false
+        setEnvVar('ELASTIC_LAYER_NAME', "elastic-apm-lambda-extension${getVersion()}")
```
v1v, line 32:
```suggestion
        setEnvVar('ELASTIC_LAYER_NAME', "elastic-apm-extension${getVersion()}")
```

.ci/Jenkinsfile
```diff
@@ -171,3 +171,15 @@ pipeline {
     }
   }
 }
+
+/**
+* Transform TAG releases from v{major}.{minor}.{patch} to
+* ver-{major}-{minor}-{patch}. f.i: given v1.2.3 then
+* ver-1-2-3.
```
v1v, line 178:
```suggestion
* -ver-1-2-3.
```

AlexanderWert:
@v1v I think we also need the architecture key as a suffix in the layer name, otherwise the ARNs will be identical for the two different architectures (which is not possible --> so would result just in two different layer versions, which is confusing)

.ci/Jenkinsfile
```diff
@@ -36,6 +35,7 @@ pipeline {
         deleteDir()
         gitCheckout(basedir: "${BASE_DIR}")
         stash allowEmpty: true, name: 'source', useDefaultExcludes: false
+        setEnvVar('ELASTIC_LAYER_NAME', "elastic-apm-extension${getVersion()}")
```
estolfo, line 38: I think we need to include a suffix for the architecture, otherwise the ARNs for x86_64 and arm64 are indistinguishable.

estolfo (changes requested):

apm-lambda-extension/Makefile
```diff
@@ -60,7 +60,7 @@ publish: validate-layer-name validate-aws-default-region
 	@aws lambda \
 		--output json \
 		publish-layer-version \
-		--layer-name "${ELASTIC_LAYER_NAME}" \
+		--layer-name "$(ELASTIC_LAYER_NAME)-$(ARCHITECTURE)" \
```
v1v, line 63: The make goal `publish` is the one in charge to apply the layer-name with the architecture since the Makefile is in charge to transform the GOARCH in the expected format.

astorm (commented):
Clearing my review queue -- a light reading of what's here looks good but I'll leave final review to @estolfo and others.

AlexanderWert (approved):

estolfo (approved):

Issue #124, opened 2022-02-10
High-level code review before GA


estolfo:
Before we GA the aws lambda extension, we should have an experienced Go engineer from another team (preferably the apm server team) review the codebase.

simitt:
@elastic/apm-server team please conduct a high level review. The goal is to agree on a GA ready state of the code. 
I suggest we put some focus points on the review:
* [ ] are errors properly handled
* [ ] any kind of deadlocks or unexpected blocking on channels; are channels closed while still consumed, etc. 
* [ ] handling of incoming data
* [ ] handling of outgoing requests to apm-server
* [ ] general data structs (allocations, references, ..)
* [ ] e2e test coverage
* [ ] unit test coverage

This is not expected to be included as go module in other programs, therefore reviewing public interfaces etc. has less priority. The `cli` is written in `nodejs`, therefore please ignore this part in the `golang` review.

estolfo:
If someone needs an overview of the code structure before doing a review, please let us know.

axw:
I've just started looking in the `logsapi` package, and how it is used, and here are some initial thoughts:

If you're not familiar with it already, I'd suggest having a look through https://github.com/golang/go/wiki/CodeReviewComments. This is followed in many Go projects, to the point that some linters have checks for the rules. You might have a bad time if you want to turn on linting later on.

Some more detailed code docs would be great. As I'm not familiar Lambda internals, or the Logs API, it wasn't clear to me what the HTTP listener was for. Linking to the relevant parts of https://docs.aws.amazon.com/lambda/latest/dg/runtimes-logs-api.html may be helpful to future readers.

Does everything that is exported need to be exported? Having a minimal public interface helps clarify which parts are implementation details, and which parts are relevant to consumers. e.g. should users of this package care about the different protocols? I realise @simitt said "This is not expected to be included as go module in other programs, therefore reviewing public interfaces etc. has less priority.", but I still think it's important for future maintainers.

The Logs API docs state:

> We also recommend setting up the local HTTP listener or the TCP port before subscribing to receive logs.

It appears that this is not being done. In main.go, subscription is happening first, and then a listener is created and finally started: https://github.com/elastic/apm-aws-lambda/blob/b010fd9a5d77d89d8b48a0e679985ee81db96a94/apm-lambda-extension/main.go#L83-L98

Is there any reason this couldn't be abstracted away? Something like:

```go
package logsapi

[...]

PR #129, opened 2022-02-22
Refactor logsapi package

estolfo:
This is a refactor of the logsapi package as per the review in #124 

Changes include:
- Expose public `Subscribe` function that subscribes to the logsapi and starts an HTTP server listening for log events
- Keep `client.go` but move the creation of the logsapi subscription structs in the client's `Subscribe` function and change the function signature
- Remove `logsapilistener.go` and move the functionality to the `Subscribe` function in `subscribe.go`
- Add `subscribe_test.go`
- Remove env variable `ELASTIC_APM_LAMBDA_LOGS_LISTENER_ADDRESS`

Todo (3 March)
- [x] Json decoder
- [ ] is it necessary to pass context to Subscribe? - See https://github.com/elastic/apm-aws-lambda/issues/143


apm-lambda-extension/logsapi/route_handlers.go
```diff
@@ -0,0 +1,69 @@
...
+		defer r.Body.Close()
+		if err != nil {
+			log.Printf("Error reading body of Logs API request: %+v", err)
+			return
+		}
+
+		var logEvents []LogEvent
+		err = json.Unmarshal(body, &logEvents)
+		if err != nil {
+			log.Println("Error unmarshalling log event batch:", err)
```
jlvoiseux, line 42:
```suggestion
			log.Printf("Error unmarshalling log event batch: %+v", err)
```

jlvoiseux (approved):
Seems great to me ! The collection of platform metrics relies on the `logsapi` package, but the only major change I see on my end is to move the definition of the platform report model from `http_listener.go` to `subscribe.go`.

apm-lambda-extension/logsapi/route_handlers.go
jlvoiseux: This small suggestion to make this print similar to the other ones used in the function.

apm-lambda-extension/logsapi/route_handlers.go
```diff
@@ -0,0 +1,69 @@
...
+	"net/http"
+
+	"github.com/pkg/errors"
+)
+
+func handleLogEventsRequest(out chan LogEvent) func(w http.ResponseWriter, r *http.Request) {
+
+	return func(w http.ResponseWriter, r *http.Request) {
+		body, err := ioutil.ReadAll(r.Body)
+		defer r.Body.Close()
```
stuartnelson3, line 33: from [the docs](https://pkg.go.dev/net/http#Request):

```golang
	// For server requests, the Request Body is always non-nil
	// but will return EOF immediately when no body is present.
	// The Server will close the request body. The ServeHTTP
	// Handler does not need to.
```

so this line can be removed

apm-lambda-extension/logsapi/route_handlers.go
```diff
@@ -0,0 +1,69 @@
...
+		if err != nil {
+			log.Printf("Error reading body of Logs API request: %+v", err)
+			return
+		}
+
+		var logEvents []LogEvent
+		err = json.Unmarshal(body, &logEvents)
```
stuartnelson3, line 40: Instead of using `ioutil.ReadAll` to get the body as a byte slice, and then using `json.Unmarshal`, you could use a `json.Decoder`:

https://pkg.go.dev/encoding/json#NewDecoder

and then you'd end up with something like

```golang
var logEvents []LogEvent
if err := json.NewDecoder(r.Body).Decode(&logEvents); err != nil {
    // handle error
}
// no error, continue on
```

apm-lambda-extension/logsapi/subscribe.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	go func() {
+		log.Printf("Extension listening for logsAPI events on %s", logsAPIListener.Addr().String())
+		logsAPIServer.Serve(logsAPIListener)
+	}()
+	return nil
+}
+
+func checkAwsSamLocal() bool {
```
stuartnelson3, line 104: go style nit: `checkAWSSamLocal`. All letters in an acronym are same case.

apm-lambda-extension/logsapi/subscribe.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+type LogEventRecord struct {
+	RequestId string `json:"requestId"`
+	Status    string `json:"status"`
+}
+
+// Subscribes to the Logs API
+func subscribe(extensionID string, eventTypes []EventType) error {
+	extensions_api_address, ok := os.LookupEnv("AWS_LAMBDA_RUNTIME_API")
```
stuartnelson3, line 51: `extensionsAPIAddress`

apm-lambda-extension/logsapi/route_handlers.go
```diff
@@ -0,0 +1,69 @@
...
+				log.Printf("Error unmarshalling log event: %+v", err)
+				w.WriteHeader(http.StatusInternalServerError)
+				continue
+			}
+			out <- logEvents[idx]
+		}
+	}
+}
+
+func (le *LogEvent) unmarshalRecord() error {
```
stuartnelson3, line 59: This sort of internal unmarshaling looks a bit non-idiomatic to me, but it's out of the scope of this PR; just leaving it here in case there's a desire to follow up in a separate PR. I would probably move this into an `UnmarshalJSON` method and handle it all there when the `LogEvent` is being originally unmarshaled

apm-lambda-extension/logsapi/subscribe.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	err = subscribe(extensionID, eventTypes)
+	if err != nil {
+		return
+	}
+	return nil
+}
+
+func startHttpServer(out chan LogEvent) (err error) {
```
stuartnelson3, line 84: `startHTTPServer`

apm-lambda-extension/logsapi/subscribe.go
```diff
@@ -0,0 +1,111 @@
...
+	mux := http.NewServeMux()
+	mux.HandleFunc("/", handleLogEventsRequest(out))
+
+	logsAPIServer = &http.Server{
+		Handler: mux,
+	}
+
+	logsAPIListener, err = net.Listen("tcp", listenerHost+":0")
+	if err != nil {
+		return
```
stuartnelson3, line 94: It's typically best practice to `return err` and not define a variable for the return types in the function signature, ie.

`startHTTPServer(out chan LogEvent) error`

apm-lambda-extension/logsapi/subscribe.go
```diff
@@ -0,0 +1,111 @@
...
+
+	go func() {
+		log.Printf("Extension listening for logsAPI events on %s", logsAPIListener.Addr().String())
+		logsAPIServer.Serve(logsAPIListener)
+	}()
+	return nil
+}
+
+func checkAwsSamLocal() bool {
+	env_aws_local, ok := os.LookupEnv("AWS_SAM_LOCAL")
```
stuartnelson3, line 105: `envAWSLocal`

apm-lambda-extension/logsapi/subscribe.go
stuartnelson3, line 84: it would probably be a good idea to pass the `context.Context` from the `Subscribe` func into this one, and listen for `<-ctx.Done()` so that you can know when to call `Shutdown()` on the `logsAPIServer`

apm-lambda-extension/logsapi/subscribe_test.go
```diff
@@ -0,0 +1,148 @@
...
+	expectedTypes := []EventType{Platform}
+	expectedBufferingCfg := BufferingCfg{
+		MaxItems:  10000,
+		MaxBytes:  262144,
+		TimeoutMS: 25,
+	}
+
+	// Create aws runtime API server and handler
+	awsRuntimeApiServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		request_bytes, _ := ioutil.ReadAll(r.Body)
```
stuartnelson3, line 61: Some general naming things in this area of code, but also another opportunity to use `json.NewDecoder`

apm-lambda-extension/logsapi/subscribe_test.go
```diff
@@ -0,0 +1,148 @@
...
+	}`
+	body := []byte(`[` + platformDoneEvent + `]`)
+	url := "http://" + logsAPIListener.Addr().String()
+	req, err := http.NewRequest("GET", url, bytes.NewReader(body))
+	if err != nil {
+		t.Log("Could not create request")
+	}
+
+	// Send the request to the logs listener
+	client := &http.Client{}
```
stuartnelson3, line 99: this is equivalent to using `http.DefaultClient` directly

apm-lambda-extension/logsapi/subscribe_test.go
```diff
@@ -0,0 +1,148 @@
...
+	go func() {
+		_, err = client.Do(req)
```
stuartnelson3, line 101: is there a reason this is being done in a go routine? The way this is written there's a race between `if err != nil` and `client.Do` executing and setting err

apm-lambda-extension/logsapi/subscribe_test.go
```diff
@@ -0,0 +1,148 @@
...
+	body := []byte(`[` + logEvent + `]`)
+	url := "http://" + logsAPIListener.Addr().String()
+	req, err := http.NewRequest("GET", url, bytes.NewReader(body))
+	if err != nil {
+		t.Log("Could not create request")
+	}
+
+	// Send the request to the logs listener
+	client := &http.Client{}
+	resp, _ := client.Do(req)
```
stuartnelson3, line 146: `http.DefaultClient`. And it's always a good idea to `require.NoError(t, err)` when you have a method/func that returns an error

stuartnelson3 (changes requested):
looking good! left some comments / questions / suggestions

apm-lambda-extension/logsapi/subscribe_test.go
estolfo: Good point about the race. It's in a go routine so that I can read from the channel in line 108. 
I could put the check for an err in the go routine and then just receive from the `out` channel, if that would make sense?

stuartnelson3: is the read from `out` synchronized with the client request returning? you could also make `out` a buffered channel, so it will store `event` without needing someone to read from it immediately, which would allow you to remove the go routine. There's still a race between the test function exiting and the `if err != nil`, since they'll be executed concurrently

apm-lambda-extension/logsapi/subscribe.go
estolfo: I thought setting the error to the variable `err` in the signature was an idiomatic way to allow you to use shorthand and just `return` in the function, as it is done in line 94. But maybe it's just a question of personal style?

apm-lambda-extension/logsapi/subscribe_test.go
estolfo: Good idea on making it a buffered channel. I think I can make this change in other tests as well.

apm-lambda-extension/logsapi/subscribe.go
estolfo: made the change anyway :)

apm-lambda-extension/logsapi/subscribe.go
stuartnelson3: My thought was that you could know when you want to exit the program, and indicate that to the server so it can allow any running requests to finish, instead of cutting them off. But maybe that's not a concern, since you'll shutdown after the lambda function has exited?

stuartnelson3 (approved):
looks good to me!

apm-lambda-extension/logsapi/subscribe.go
estolfo: I'll move this into another issue because I'm not sure what the best approach is here and we are waiting on the other changes in this PR.
Issue: https://github.com/elastic/apm-aws-lambda/issues/143

PR #132, opened 2022-02-23
Address partial list of code review comments

estolfo:
This PR addresses some comments from this code review: https://github.com/elastic/apm-aws-lambda/issues/124#issuecomment-1042104258

Todo (3 March)
- [x] Use JSON.Decoder

### DIFF
--- a/apm-lambda-extension/extension/client.go
+++ b/apm-lambda-extension/extension/client.go
@@ -22,7 +22,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"log"
 	"net/http"
 )
@@ -110,13 +109,8 @@ func (e *Client) Register(ctx context.Context, filename string) (*RegisterRespon
 	if httpRes.StatusCode != 200 {
 		return nil, fmt.Errorf("extension register request failed with status %s", httpRes.Status)
 	}
-	body, err := ioutil.ReadAll(httpRes.Body)
-	if err != nil {
-		return nil, err
-	}
 	res := RegisterResponse{}
-	err = json.Unmarshal(body, &res)
-	if err != nil {
+	if err := json.NewDecoder(httpRes.Body).Decode(&res); err != nil {
 		return nil, err
 	}
 	e.ExtensionID = httpRes.Header.Get(extensionIdentiferHeader)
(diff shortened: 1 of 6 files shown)


apm-lambda-extension/extension/client.go
```diff
@@ -105,10 +105,11 @@ func (e *Client) Register(ctx context.Context, filename string) (*RegisterRespon
 	if err != nil {
 		return nil, err
 	}
+	defer httpRes.Body.Close()
+
 	if httpRes.StatusCode != 200 {
 		return nil, fmt.Errorf("extension register request failed with status %s", httpRes.Status)
 	}
-	defer httpRes.Body.Close()
 	body, err := ioutil.ReadAll(httpRes.Body)
```
stuartnelson3, line 113: here's another opportunity to use `json.Decoder`

apm-lambda-extension/extension/http_server_test.go
```diff
@@ -168,6 +169,24 @@ func Test_handleInfoRequest(t *testing.T) {
 	}
 }
 
+type errReader int
+
+func (errReader) Read(p []byte) (n int, err error) {
```
stuartnelson3, line 174: nit, if you're not using any of the variables you could change the method sig to
```golang
func (errReader) Read(_ []byte) (int, error)
```

apm-lambda-extension/extension/route_handlers.go
```diff
@@ -37,24 +37,26 @@ func handleInfoRequest(apmServerUrl string) func(w http.ResponseWriter, r *http.
 		client := &http.Client{}
 
 		req, err := http.NewRequest(r.Method, apmServerUrl, nil)
+		if err != nil {
+			log.Printf("could not create request object for %s:%s: %v", r.Method, apmServerUrl, err)
+			return
+		}
```
stuartnelson3, line 43: Since this is happening within a handler, it would be good to inform the client that the server had an error. I think this early return will result in the client receiving a 200.

stuartnelson3 (approved):
looks good! made some comments but i think they're mostly cosmetic

trentm:
FYI: When testing with a build of the extension using this branch and with the Node.js APM agent changes to send the `?flushed=true` signal (https://github.com/elastic/apm-agent-nodejs/pull/2586), I no longer get the panics I reported in https://github.com/elastic/apm-aws-lambda/issues/133

estolfo:
Thanks for the update @trentm. [This](https://github.com/elastic/apm-aws-lambda/pull/132/files#diff-266b2e81a44ec5888932f7d39596edd7855613da2a972f75b75eec6f266ad363L204-L205) change should have indeed fixed that panic.
Also, @jlvoiseux has written a test in [this PR](https://github.com/elastic/apm-aws-lambda/pull/136) that forces the scenario and confirms that the changes linked above resolve the panic.

stuartnelson3 (approved):
looks good!

Issue #133, opened 2022-02-24
"panic: send on closed channel" from the extension


trentm:
This is the tail end of a AWS Lambda log for a small Node.js function execution. It shows a panic in the extension:

```
...
2022-02-23T23:00:11.485000+00:00 2022/02/23/[$LATEST]743a0f5ca5744087a061a92d7cb56e5e {"log.level":"trace","@timestamp":"2022-02-23T23:00:11.485Z","log":{"logger":"elastic-apm-node"},"event.module":"apmclient","reqId":"e3f64a016331cdefab5ad1fe46b854d2","timeline":[[219.337168,"completePart gzipStream",null],[219.811828,"completePart intakeReq",null],[357.343887,"completePart intakeRes",null]],"bytesWritten":800,"backoffDelayMs":0,"ecs":{"version":"1.6.0"},"message":"conclude intake request: success"}
2022-02-23T23:00:11.507000+00:00 2022/02/23/[$LATEST]743a0f5ca5744087a061a92d7cb56e5e 2022-02-23T23:00:11.485Z	e3918d8f-7391-4bea-a6e2-9fdefc4c9037	INFO	XXX onFlushed in _writeFlush called
2022-02-23T23:00:11.524000+00:00 2022/02/23/[$LATEST]743a0f5ca5744087a061a92d7cb56e5e 2022-02-23T23:00:11.524Z	e3918d8f-7391-4bea-a6e2-9fdefc4c9037	INFO	XXX _signalLambdaEnd start
2022-02-23T23:00:11.526000+00:00 2022/02/23/[$LATEST]743a0f5ca5744087a061a92d7cb56e5e {"log.level":"trace","@timestamp":"2022-02-23T23:00:11.525Z","log":{"logger":"elastic-apm-node"},"awsRequestId":"e3918d8f-7391-4bea-a6e2-9fdefc4c9037","ecs":{"version":"1.6.0"},"message":"lambda: flushed"}
2022-02-23T23:00:11.565000+00:00 2022/02/23/[$LATEST]743a0f5ca5744087a061a92d7cb56e5e 2022/02/23 23:00:11 Received agent done signal
[...]

Issue #128, opened 2022-02-21
panic: sync: WaitGroup is reused before previous Wait has returned


estolfo:
Error from the extension:
```
panic: sync: WaitGroup is reused before previous Wait has returned
goroutine 1 [running]:
sync.(*WaitGroup).Wait(0xc0001a3f10)
/usr/local/go/src/sync/waitgroup.go:132 +0xa5
main.main()
main.go:197 +0x87e
```

We'll have to revisit our use of a waitgroup to allow the send to the apm server to complete before the extension is frozen. Added in https://github.com/elastic/apm-aws-lambda/pull/106

jlvoiseux:
## Current reproduction results

### Repro 1 : `main` branch, Ruby Runtime, function [EmilyTestRepro](https://eu-central-1.console.aws.amazon.com/lambda/home?region=eu-central-1#/functions/EmilyTestRepro?tab=testing)

With a lambda function similar to the one used by Emily, the error is successfully reproduced. I added additional to the extension to monitor the behavior of `backgroundDataSendWg`, which is the `WaitGroup` triggering the panic. The complete [logs](https://eu-central-1.console.aws.amazon.com/cloudwatch/home?region=eu-central-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252FEmilyTestRepro) are available here. The most relevant excerpt is below : 
```
START RequestId: 757912f1-8f13-4a2e-bb3a-a28619dd6b28 Version: $LATEST
2022/02/24 13:26:40 546ab00f-33b1-4915-ad6b-bb8d45ca5f37
2022/02/24 13:26:40 Register response: {
	"functionName": "EmilyTestRepro",
	"functionVersion": "$LATEST",
	"handler": "lambda_function.lambda_handler"
}
2022/02/24 13:26:40 Could not read ELASTIC_APM_DATA_RECEIVER_TIMEOUT_SECONDS, defaulting to 15: strconv.Atoi: parsing "": invalid syntax
2022/02/24 13:26:40 Extension listening for apm data on :8200
LOGS	Name: extension-dev-linux-amd64	State: Subscribed	Types: [platform]
2022/02/24 13:26:40 Waiting for next event...
2022/02/24 13:26:40 Server listening for logs data from AWS Logs API on sandbox:1234
[...]

PR #136, opened 2022-03-01
Add unit tests for main.go

jlvoiseux:
### Motivation
The goal of these tests is to ease the refactor of the extension following the recent code reviews. They can also be used to solve some of the current issues. For example : 
- `TestFlush` reproduces #133 (a `RuntimeDone` signal is sent to the extension after the closure of the channel
- `TestSeveralTransactions` reproduces #128 (race condition causing a `WaitGroup` to be modified while it hangs)

The core idea is to combine a mock Lambda API backend with a mock APM server to enable the recreation of several scenarios.

### Limitations 
- Since #132 is not merged yet, `TestFlush` panics. Same for `TestSeveralTransactions`, as the related issue is not yet solved.
- Another limitation is the fact that due to the concurrent nature of the processes tested, the aforementioned failing tests might sometimes pass.

### DIFF
--- a/apm-lambda-extension/main_test.go
+++ b/apm-lambda-extension/main_test.go
@@ -1,0 +1,206 @@
+package main
+
+import (
+	"bytes"
+	"elastic/apm-lambda-extension/extension"
+	"elastic/apm-lambda-extension/logsapi"
+	json "encoding/json"
+	"github.com/google/uuid"
+	"log"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"strings"
+	"testing"
+	"time"
+)
+
+type RegistrationResponse struct {
+	FunctionName    string `json:"functionName"`
+	FunctionVersion string `json:"functionVersion"`
+	Handler         string `json:"handler"`
+}
+
+func initMockServers(eventsChannel chan MockEvent) (*httptest.Server, *httptest.Server) {
+
+	// Mock APM Server
+	APMServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.RequestURI == "/intake/v2/events" {
+			w.WriteHeader(http.StatusAccepted)
+		}
+	}))
+	os.Setenv("ELASTIC_APM_LAMBDA_APM_SERVER", APMServer.URL)
+	os.Setenv("ELASTIC_APM_SECRET_TOKEN", "none")
+
+	// Mock Lambda Server
+	lambdaServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch r.RequestURI {
+		// Extension registration request
+		case "/2020-01-01/extension/register":
+			w.Header().Set("Lambda-Extension-Identifier", "b03a29ec-ee63-44cd-8e53-3987a8e8aa8e")
+			body, _ := json.Marshal(RegistrationResponse{
+				FunctionName:    "UnitTestingMockLambda",
+				FunctionVersion: "$LATEST",
+				Handler:         "main_test.mock_lambda",
+			})
+			w.Write(body)
+		case "/2020-01-01/extension/event/next":
+			currId := uuid.New().String()
+			select {
+			case nextEvent := <-eventsChannel:
+				w.Write(sendNextEventInfo(currId, nextEvent))
+				go processMockEvent(currId, nextEvent, APMServer)
+			default:
+				finalShutDown := MockEvent{
+					Type:              Shutdown,
+					Name:              "finalShutdown",
+					ExecutionDuration: 0,
+					Timeout:           0,
+				}
+				w.Write(sendNextEventInfo(currId, finalShutDown))
+			}
+		// Logs API subscription request
+		case "/2020-08-15/logs":
+			w.WriteHeader(http.StatusOK)
+			os.Setenv("ELASTIC_APM_LAMBDA_LOGS_LISTENER_ADDRESS", "localhost:8205")
+		}
+	}))
+
+	slicedLambdaURL := strings.Split(lambdaServer.URL, "//")
+	strippedLambdaURL := slicedLambdaURL[1]
+	os.Setenv("AWS_LAMBDA_RUNTIME_API", strippedLambdaURL)
+	extensionClient = extension.NewClient(os.Getenv("AWS_LAMBDA_RUNTIME_API"))
+
+	return lambdaServer, APMServer
+}
+
+type MockEventType string
+
+const (
+	InvokeHang                 MockEventType = "Hang"
+	InvokeStandard             MockEventType = "Standard"
+	InvokeStandardFlush        MockEventType = "Flush"
+	InvokeMultipleTransactions MockEventType = "MultipleTransactions"
+	Shutdown                   MockEventType = "Shutdown"
+)
+
+type MockEvent struct {
+	Type              MockEventType
+	Name              string
+	ExecutionDuration float64
+	Timeout           float64
+}
+
+func processMockEvent(currId string, event MockEvent, APMServer *httptest.Server) {
+	sendLogEvent(currId, "platform.start")
+	client := http.Client{}
+	switch event.Type {
+	case InvokeHang:
+		time.Sleep(time.Duration(event.Timeout) * time.Second)
+	case InvokeStandard:
+		time.Sleep(time.Duration(event.ExecutionDuration) * time.Second)
+		req, _ := http.NewRequest("POST", "http://localhost:8200/intake/v2/events", bytes.NewBuffer([]byte(event.Name)))
+		client.Do(req)
+	case InvokeStandardFlush:
+		time.Sleep(time.Duration(event.ExecutionDuration) * time.Second)
+		reqData, _ := http.NewRequest("POST", "http://localhost:8200/intake/v2/events?flushed=true", bytes.NewBuffer([]byte(event.Name)))
+		client.Do(reqData)
+	case InvokeMultipleTransactions:
+		time.Sleep(time.Duration(event.ExecutionDuration) * time.Second)
+		reqData0, _ := http.NewRequest("POST", "http://localhost:8200/intake/v2/events", bytes.NewBuffer([]byte(event.Name)))
+		reqData1, _ := http.NewRequest("POST", "http://localhost:8200/intake/v2/events", bytes.NewBuffer([]byte(event.Name+"2")))
+		go client.Do(reqData0)
+		go client.Do(reqData1)
+		time.Sleep(650 * time.Microsecond)
+	}
+	sendLogEvent(currId, "platform.runtimeDone")
+}
+
+func sendNextEventInfo(id string, event MockEvent) []byte {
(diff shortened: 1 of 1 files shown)


apm-lambda-extension/e2e-testing/e2e_test.go
```diff
@@ -1,11 +1,6 @@
 package e2e_testing
```
jlvoiseux, line 1: I restructured the e2e testing code in order to make some utility functions (decompress, unzip, run bash command, ...) available to other packages.

apm-lambda-extension/extension/process_env.go
```diff
@@ -25,12 +25,13 @@ import (
 )
```
jlvoiseux, line 25: Addition of the timeout env variable. It is in uppercase in the `extensionConfig` struct because it needs to be exported. In the end, I chose to keep this change into the PR, as it is required for all the Unit Tests to pass.

apm-lambda-extension/main_test.go
```diff
@@ -0,0 +1,343 @@
+package main
```
jlvoiseux, line 1: This is the Unit testing framework itself.

apm-lambda-extension/extension/process_env.go
```diff
@@ -60,8 +61,14 @@ func getIntFromEnv(name string) (int, error) {
...
 	dataReceiverTimeoutSeconds, err := getIntFromEnv("ELASTIC_APM_DATA_RECEIVER_TIMEOUT_SECONDS")
 	if err != nil {
-		log.Printf("Could not read ELASTIC_APM_DATA_RECEIVER_TIMEOUT_SECONDS, defaulting to 15: %v\n", err)
 		dataReceiverTimeoutSeconds = 15
+		log.Printf("Could not read ELASTIC_APM_DATA_RECEIVER_TIMEOUT_SECONDS, defaulting to %d: %v\n", dataReceiverTimeoutSeconds, err)
+	}
+
+	dataForwarderTimeoutSeconds, err := getIntFromEnv("ELASTIC_APM_DATA_FORWARDER_TIMEOUT_SECONDS")
+	if err != nil {
+		dataForwarderTimeoutSeconds = 3
```
estolfo, line 70: Maybe we should have these defaults as constants? Otherwise we'll have to search in the code for the place where we set the default.

apm-lambda-extension/extension/route_handlers.go
```diff
@@ -99,8 +99,13 @@ func handleIntakeV2Events(agentDataChan chan AgentData) func(w http.ResponseWrit
...
 				ContentEncoding: r.Header.Get("Content-Encoding"),
 			}
-			log.Println("Adding agent data to buffer to be sent to apm server")
-			agentDataChan <- agentData
+
+			select {
+			case agentDataChan <- agentData:
+				log.Println("Adding agent data to buffer to be sent to apm server")
+			default:
+				log.Println("Channel full : dropping event")
```
estolfo, line 107:
```suggestion
				log.Println("Channel full: dropping event")
```

apm-lambda-extension/main.go
```diff
@@ -121,17 +122,17 @@ func main() {
...
 			// This is usually due to inactivity.
 			if event.EventType == extension.Shutdown {
 				extension.ProcessShutdown()
 				return
 			}
 
+			// Flush any APM data, in case waiting for the agentDone or runtimeDone signals
+			// timed out, the agent data wasn't available yet, and we got to the next non-shutdown event
+			extension.FlushAPMData(client, agentDataChannel, config)
+
```
estolfo, line 135: Why did you move this block to after the Shutdown event check?

apm-lambda-extension/main_test.go
```diff
@@ -0,0 +1,343 @@
...
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"strings"
+	"sync"
+	"testing"
+	"time"
+)
+
+type RegistrationResponse struct {
```
estolfo, line 20: We already have a struct for this extension.RegisterResponse, maybe you can reuse it?

apm-lambda-extension/main_test.go
```diff
@@ -0,0 +1,343 @@
...
+		DeadlineMs:         time.Now().UnixNano()/int64(time.Millisecond) + int64(event.Timeout*1000),
+		RequestID:          id,
+		InvokedFunctionArn: "arn:aws:lambda:eu-central-1:627286350134:function:main_unit_test",
+		Tracing:            extension.Tracing{},
+	}
+	if event.Type == Shutdown {
+		nextEventInfo.EventType = "SHUTDOWN"
+	}
+
+	out, _ := json.Marshal(nextEventInfo)
```
estolfo, line 176: can we use [`json.Encoder`](https://pkg.go.dev/encoding/json#Encoder) here? stuart was suggesting we use `json.Decoder` instead of `Unmarshal` directly.

apm-lambda-extension/main_test.go
```diff
@@ -0,0 +1,343 @@
...
+	record := logsapi.LogEventRecord{
+		RequestId: requestId,
+	}
+	logEvent := logsapi.LogEvent{
+		Time:   time.Now(),
+		Type:   logEventType,
+		Record: record,
+	}
+	logEvent.RawRecord, _ = json.Marshal(logEvent.Record)
+	body, _ := json.Marshal([]logsapi.LogEvent{logEvent})
```
estolfo, line 190: same here, maybe we can use `json.Encoder`

apm-lambda-extension/main_test.go
```diff
@@ -0,0 +1,343 @@
...
+}
+
+// TESTS
+
+func TestStandardEventsChain(t *testing.T) {
+	http.DefaultServeMux = new(http.ServeMux)
+	log.Println("Standard Test")
+
+	eventsChannel := make(chan MockEvent, 100)
+	defer close(eventsChannel)
```
estolfo, line 209: we probably don't need to close `eventsChannel` explicitly, unless there's a reason to?

estolfo (changes requested):
This is really great, thanks for building this test framework! I'm just wondering if there's some way we can add an assertion to the tests or if it makes sense to have the absence of a panic indicate that the test passes.

apm-lambda-extension/logsapi/subscribe.go
```diff
@@ -29,9 +29,9 @@ import (
 	"github.com/pkg/errors"
```
jlvoiseux, line 29: I exposed `listenerHost`, `logsAPIServer` and `logsAPIListener` for testing purposes. The linter recommended against using the package name in variable names, hence the renaming of `logsAPIServer` and `LogAPIListener`

apm-lambda-extension/main.go
jlvoiseux: When writing my tests, I found myself in the following situation : 

Test setup : 
- APM Server configured to be very slow (approx 3 seconds to respond).
- APM Agent Data channel filled with data
- SendStrategy configured to background

Behavior :
The extension starts sending data from the loaded channel to the slow APM server. It is then interrupted due to the current event being finished. A shutdown event is then received, and we start flushing APM data. The issue is that this takes a tremendous time if the APM Agent data channel is full and if the response time of the Server is large, even if the SendStrategy is set to `background`.

An alternative solution would be to put the flush statement above the Shutdown, but to only run `FlushAPMData` if the SendStrategy is `SyncFlush`

apm-lambda-extension/extension/process_env.go
jlvoiseux: Solved in [d9a8edc](https://github.com/elastic/apm-aws-lambda/pull/136/commits/d9a8edc0ca29cd924a0217ed22f9715a5c270c90)

apm-lambda-extension/extension/route_handlers.go
jlvoiseux: Solved in [d9a8edc](https://github.com/elastic/apm-aws-lambda/pull/136/commits/d9a8edc0ca29cd924a0217ed22f9715a5c270c90)

apm-lambda-extension/main_test.go
```diff
@@ -0,0 +1,343 @@
...
+	FunctionName    string `json:"functionName"`
+	FunctionVersion string `json:"functionVersion"`
+	Handler         string `json:"handler"`
+}
+
+func initMockServers(eventsChannel chan MockEvent) (*httptest.Server, *httptest.Server, chan struct{}) {
+
+	// Mock APM Server
+	hangChan := make(chan struct{})
+	APMServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
jlvoiseux: Indeed ! Solved in [d9a8edc](https://github.com/elastic/apm-aws-lambda/pull/136/commits/d9a8edc0ca29cd924a0217ed22f9715a5c270c90)

apm-lambda-extension/main_test.go
jlvoiseux: Solved in [d9a8edc](https://github.com/elastic/apm-aws-lambda/pull/136/commits/d9a8edc0ca29cd924a0217ed22f9715a5c270c90)

apm-lambda-extension/main_test.go
jlvoiseux: Solved in [d9a8edc](https://github.com/elastic/apm-aws-lambda/pull/136/commits/d9a8edc0ca29cd924a0217ed22f9715a5c270c90)

apm-lambda-extension/main_test.go
jlvoiseux: Solved in [d9a8edc](https://github.com/elastic/apm-aws-lambda/pull/136/commits/d9a8edc0ca29cd924a0217ed22f9715a5c270c90)

apm-lambda-extension/main_test.go
jlvoiseux: Indeed, solved in [d9a8edc](https://github.com/elastic/apm-aws-lambda/pull/136/commits/d9a8edc0ca29cd924a0217ed22f9715a5c270c90)

jlvoiseux:
I added success/failure assertions in [41a8039](https://github.com/elastic/apm-aws-lambda/pull/136/commits/41a803961c6e851049248bcfd11d9b063aab4052). For most tests, I check if the extension does not panic and if APM data is correctly sent to the APM server.

apm-lambda-extension/main.go
estolfo: Thanks for the explanation. I think we might want to remove this flush anyway: with `syncFlush`, we'll flush at the end of an invocation so there won't be anything to flush upon a new function invocation. With `background`, we'll just continue receiving from the channel in the go routine when the extension starts its work for a new function invocation.

estolfo (approved):
👍🏼

estolfo (approved):
Just two minor comments then I think we can merge this. Thanks for your work on this!

jlvoiseux:
@estolfo Thank you for the review ! Your two suggestions have been addressed in [0f58cb5](https://github.com/elastic/apm-aws-lambda/pull/136/commits/0f58cb5954ac3541dd72c21a916768c6a70efcca)

PR #139, opened 2022-03-03
doc: update node.js docs to mention layers

astorm:
Updates the docs to point users at the Node Agents layer.

### DIFF
--- a/docs/aws-lambda-extension.asciidoc
+++ b/docs/aws-lambda-extension.asciidoc
@@ -118,7 +118,14 @@ You'll then need to configure your function to use that layer. To add a layer
 4. Enter the Version ARN of the APM Lambda Extension layer in the _Specify an ARN_ text input
 5. Click the _Add_ button
 
-The Node.js agent will wrap your Lambda function handler automatically. You can also use the following syntax to manually wrap the handler.
+Finally, to have the the Node.js automatically wrap your Lambda function handler, just add the following environment variable.
+
+[source]
+---
+NODE_OPTIONS=-r elastic-apm-node/start
+---
+
+You can also use the following syntax to manually wrap the handler.
 
 [source,js]
 ----
--- a/docs/aws-lambda-extension.asciidoc
+++ b/docs/aws-lambda-extension.asciidoc
@@ -125,20 +125,6 @@ Finally, to have the Node.js agent automatically wrap your Lambda function handl
 NODE_OPTIONS=-r elastic-apm-node/start
 ---
 
-You can also use the following syntax to manually wrap the handler.
-
-[source,js]
-----
-const apm = require('elastic-apm-node').start({/*...*/})
-exports.handler = apm.lambda(async function handler (event, context) {
-    const response = {
-      statusCode: 200,
-      body: "hello new async."
-    };
-    return response
-})
-----
-
 See the {apm-node-ref}/lambda.html[Node.js agent setup guide] for detailed instructions on setting up the Node.js agent for AWS Lambda.
 
 [discrete]
--- a/docs/aws-lambda-extension.asciidoc
+++ b/docs/aws-lambda-extension.asciidoc
@@ -98,7 +98,7 @@ For a _minimal configuration_ you need to specify the `ELASTIC_APM_LAMBDA_APM_SE
 [[aws-lambda-handler]]
 === Configuring the Agent and Lambda Function handler
 
-Once you've installed the extension, there's one last step to take. You'll need to install the correct language agent and wrap the Lambda function handler.
+After the extension has been installed, install the relevant language agent and wrap the Lambda function handler.
 
 
 [discrete]
--- a/docs/aws-lambda-extension.asciidoc
+++ b/docs/aws-lambda-extension.asciidoc
@@ -105,7 +105,7 @@ Once you've installed the extension, there's one last step to take. You'll need 
 [[aws-lambda-nodejs]]
 ==== Node.js
 
-To install the Node.js agent in Lambda you'll use a second layer.  You'll need to https://github.com/elastic/apm-agent-nodejs/releases[pick the right ARN from the agent release table]. The ARN has the pattern `arn:aws:lambda:<AWS_REGION_KEY>:267093732750:layer:elastic-apm-node-ver-<APM_EXTENSION_VERSION>-0:<LAYER_VERSION>` and depends on:
+To install the Node.js agent in Lambda you'll use a second layer.  You'll need to https://github.com/elastic/apm-agent-nodejs/releases[pick the right ARN from the agent release table]. The ARN has the pattern `arn:aws:lambda:<AWS_REGION_KEY>:267093732750:layer:elastic-apm-node-ver-<APM_EXTENSION_VERSION>:<LAYER_VERSION>` and depends on:
 
 * The AWS region your Lambda function runs in. The APM Lambda Extension layer needs to be in the same region as your Lambda function.
 * The version of the APM Lambda Extension you would like to use.
--- a/docs/aws-lambda-extension.asciidoc
+++ b/docs/aws-lambda-extension.asciidoc
@@ -118,7 +118,7 @@ You'll then need to configure your function to use that layer. To add a layer
 4. Enter the Version ARN of the APM Lambda Extension layer in the _Specify an ARN_ text input
 5. Click the _Add_ button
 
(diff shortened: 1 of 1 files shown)


astorm:
@bmorelli25 From a tech writing/elastic standards point of view, do you think we need to repeat the "how to add a layer in the UI" instructions a second time (as its written now), or can we assume the user will have picked this up from the initial section?

docs/aws-lambda-extension.asciidoc
```diff
@@ -98,13 +98,34 @@ For a _minimal configuration_ you need to specify the `ELASTIC_APM_LAMBDA_APM_SE
...
-Once you've installed the extension, there's one last step to take. You'll need to wrap the Lambda function handler.
+Once you've installed the extension, there's one last step to take. You'll need to install the correct language agent and wrap the Lambda function handler.
+
 
 [discrete]
 [[aws-lambda-nodejs]]
 ==== Node.js
 
-In Node.js, you wrap a Lambda function handler using the following syntax.
+To install the Node.js agent in Lambda you'll use a second layer.  You'll need to https://github.com/elastic/apm-agent-nodejs/releases[pick the right ARN from the agent release table]. The ARN has the pattern `arn:aws:lambda:<AWS_REGION_KEY>:267093732750:layer:elastic-apm-node-ver-<APM_EXTENSION_VERSION>-0:<LAYER_VERSION>` and depends on:
```
trentm, line 108:
```suggestion
To install the Node.js agent in Lambda you'll use a second layer.  You'll need to https://github.com/elastic/apm-agent-nodejs/releases[pick the right ARN from the agent release table]. The ARN has the pattern `arn:aws:lambda:<AWS_REGION_KEY>:267093732750:layer:elastic-apm-node-ver-<APM_EXTENSION_VERSION>:<LAYER_VERSION>` and depends on:
```

docs/aws-lambda-extension.asciidoc
```diff
@@ -98,13 +98,34 @@ For a _minimal configuration_ you need to specify the `ELASTIC_APM_LAMBDA_APM_SE
...
+5. Click the _Add_ button
+
+Finally, to have the the Node.js automatically wrap your Lambda function handler, just add the following environment variable.
+
+[source]
+---
+NODE_OPTIONS=-r elastic-apm-node/start
+---
+
+You can also use the following syntax to manually wrap the handler.
```
trentm, line 128: I think users might be confused by this last sentence on whether they *need* to manually `apm.lambda()` wrap now and whether they should.

What about removing this block on manual wrapping entirely from this doc, because it is no longer the blessed setup method, and moving discussion of manual wrapping to a subsection of the Node.js Agent-specific docs?


I notice that the Java section below doesn't include details on installing the Java Agent layer in this doc, but instead just points to the Java-Agent specific doc. Thoughts on doing the same? I don't have a strong opinion.

docs/aws-lambda-extension.asciidoc
```diff
@@ -98,13 +98,34 @@ For a _minimal configuration_ you need to specify the `ELASTIC_APM_LAMBDA_APM_SE
...
+
+You'll then need to configure your function to use that layer. To add a layer
+
+1. Navigate to your function in the AWS Console
+2. Scroll to the Layers section and click the _Add a layer_ button image:images/config-layer.png[image of layer configuration section in AWS Console]
+3. Choose the _Specify an ARN_ radio button
+4. Enter the Version ARN of the APM Lambda Extension layer in the _Specify an ARN_ text input
+5. Click the _Add_ button
+
+Finally, to have the the Node.js automatically wrap your Lambda function handler, just add the following environment variable.
```
bmorelli25, line 121:
```suggestion
Finally, to have the Node.js automatically wrap your Lambda function handler, add the following environment variable.
```

docs/aws-lambda-extension.asciidoc
```diff
@@ -98,13 +98,34 @@ For a _minimal configuration_ you need to specify the `ELASTIC_APM_LAMBDA_APM_SE
 [[aws-lambda-handler]]
 === Configuring the Agent and Lambda Function handler
 
-Once you've installed the extension, there's one last step to take. You'll need to wrap the Lambda function handler.
+Once you've installed the extension, there's one last step to take. You'll need to install the correct language agent and wrap the Lambda function handler.
```
bmorelli25, line 101: This feels like a bit of a lie as "one last step" is really six more steps. What about this instead?

```suggestion
After the extension has been installed, install the relevant language agent and wrap the Lambda function handler.
```

docs/aws-lambda-extension.asciidoc
```diff
@@ -98,13 +98,34 @@ For a _minimal configuration_ you need to specify the `ELASTIC_APM_LAMBDA_APM_SE
...
+* The AWS region your Lambda function runs in. The APM Lambda Extension layer needs to be in the same region as your Lambda function.
+* The version of the APM Lambda Extension you would like to use.
+
+You'll then need to configure your function to use that layer. To add a layer
+
+1. Navigate to your function in the AWS Console
+2. Scroll to the Layers section and click the _Add a layer_ button image:images/config-layer.png[image of layer configuration section in AWS Console]
+3. Choose the _Specify an ARN_ radio button
+4. Enter the Version ARN of the APM Lambda Extension layer in the _Specify an ARN_ text input
+5. Click the _Add_ button
```
bmorelli25, line 119: > From a tech writing/elastic standards point of view, do you think we need to repeat the "how to add a layer in the UI" instructions a second time (as its written now), or can we assume the user will have picked this up from the initial section?

This is hard for me to answer without a solid understanding of AWS Lambda. If I'm understanding correctly, this section of text is duplicated from above, and the Node.js agent is the only agent that needs to run a second layer? My gut is to say yes, it makes sense to repeat these instructions. If they're always going to be the same as the instructions above, we can single-source them by "tagging" the text above and including it down here. That will help with maintenance.

docs/aws-lambda-extension.asciidoc
AlexanderWert: > I notice that the Java section below doesn't include details on installing the Java Agent layer in this doc, but instead just points to the Java-Agent specific doc. Thoughts on doing the same? I don't have a strong opinion.

+1 on this proposal. We need a meaningful docs page for Lambda within the Node.js agent docs as well. However, if we describe the Node-specific setup in detail here, then we would need to duplicate the docs.
So I'd propose to follow the splitting of Lambda docs in the way, that on this docs page we have a clear focus on setting up the extension plus clear references to agent-specific docs, but having the language / agent-specific docs on the corresponding agent docs pages.

AlexanderWert (changes requested):

docs/aws-lambda-extension.asciidoc
estolfo: I agree with this

estolfo (approved):

docs/aws-lambda-extension.asciidoc
astorm: ✅

docs/aws-lambda-extension.asciidoc
astorm: I hear two things in the feedback.  The first is that we should remove this section on wrapping the handler manually for the Node Agent. I've done this in  80c9342272a2eb28441a263faa6524ce214a42e1

The second is more vague and less actionable -- 

> We need a meaningful docs page for Lambda within the Node.js agent docs as well
> ...
>  So I'd propose to follow the splitting of Lambda docs in the way, that on this docs page we have a clear focus on setting up the extension plus clear references to agent-specific docs, but having the language / agent-specific docs on the corresponding agent docs pages.

I'm unsure if this is just confirming that we want to keep a link to the Node Agent's lambda page, or if there's a request for additional changes to the Node Agent's lambda page.  There doesn't appear to be any additional action needed for this doc.  If that's incorrect please let me know what needs to be done.

AlexanderWert: > I'm unsure if this is just confirming that we want to keep a link to the Node Agent's

@astorm sorry, yes, this was a bunch of fuzzy words from my side to actually just confirm the link to Node Agent's docs :-).

Regarding the docs in the Node agent: is [this step](https://github.com/elastic/apm-agent-nodejs/blob/main/docs/lambda.asciidoc#instrumenting-your-lambda-handler) (adding `elastic-apm-node` as dependency) still needed with the Node wrapper in place? (but we can continue the discussion on this aspect somewhere else, it's not affecting this PR)

AlexanderWert (approved):

Issue #131, opened 2022-02-23
Test and handle case when APM server doesn't respond


estolfo:
We need to handle the case when the APM server doesn't respond to extension requests. The extension should never hold up the runtime so we'll opt for dropping the APM data that failed to send instead of retrying.

See issue #130 for a possible related example

felixbarny:
When the queue is full, we should still send a success status code to the agents, even if we end up dropping the data. Otherwise, the agent will do the [exponential backoff](https://github.com/elastic/apm/blob/main/specs/agents/transport.md#transport-errors) and start buffering events in its internal queue. This can lead to the agent being unable to enqueue and process flush events, therefore blocking the lambda.

jlvoiseux:
Work towards closing this issue has been done in [c7384e0](https://github.com/elastic/apm-aws-lambda/pull/136/commits/c7384e08f52c3aa0c8612c2bc98b9c5f6ebc1c00). Modifications include: 

- Addition of a Timeout config option for the HTTP client performing APM queries (see issue):
https://github.com/jlvoiseux/apm-aws-lambda/blob/c7384e08f52c3aa0c8612c2bc98b9c5f6ebc1c00/apm-lambda-extension/extension/process_env.go#L68-L72
https://github.com/jlvoiseux/apm-aws-lambda/blob/c7384e08f52c3aa0c8612c2bc98b9c5f6ebc1c00/apm-lambda-extension/main.go#L71-L74

- Drop events when the channel is full:
https://github.com/jlvoiseux/apm-aws-lambda/blob/c7384e08f52c3aa0c8612c2bc98b9c5f6ebc1c00/apm-lambda-extension/extension/route_handlers.go#L102-L107

jlvoiseux:
Following discussions with @felixbarny and @estolfo, we decided to implement the following backoff mechanism:
- Upon detection of a sending error in `extension.PostToAPMServer()` (used in both `background` and `syncflush` strategies) we enter a backoff state.
- Upon backoff state entry, we start to wait for an amount of time specified in the [Transport specification](https://github.com/elastic/apm/blob/main/specs/agents/transport.md#transport-errors). During this time interval, nothing is sent to the APM server, either in `background` or through the `FlushAPMData()` function enabled by the `syncflush` strategy.
[...]

PR #148, opened 2022-03-15
Add backoff mechanism to recover from APM Server transport failures

jlvoiseux:
### Motivation / Summary
This PR aims to implement a backoff mechanism to recover from APM Server transport failures, due to the APM server being unavailable. To meet this goal, the mechanism described in issue #131 is implemented.

### Changes
In `apm_server.go`, a simple finiste state machine is implemented. The states are as follows : 
- `nominal` : in this state, the extension is able to send APM data without any issue.
- `failure` : we enter this state upon detection of a transport error. In this state, no data is sent to the APM server (the data is queued back into the dedicated channel).
- `waitingForNextAttempt` : upon detection of a transport error, a timer (defined according to the APM Transport [spec](https://github.com/elastic/apm/blob/main/specs/agents/transport.md#transport-errors)) is started. When it ends, this state is used to signal to the extension that it can try again to send data to the APM server.#147 

### Pending questions
- [ ] When should we stop trying to reconnect to the APM server ? How can we relate this mechanism to the `ELASTIC_APM_DATA_FORWARDER_TIMEOUT_SECONDS` config option.

### DIFF
--- a/apm-lambda-extension/extension/apm_server.go
+++ b/apm-lambda-extension/extension/apm_server.go
@@ -24,13 +24,32 @@ import (
 	"io"
 	"io/ioutil"
 	"log"
+	"math"
+	"math/rand"
 	"net/http"
 	"sync"
+	"time"
 )
 
 var bufferPool = sync.Pool{New: func() interface{} {
 	return &bytes.Buffer{}
 }}
+
+type APMServerTransportStatusType string
+
+const (
+	failure               APMServerTransportStatusType = "failure"
+	waitingForNextAttempt APMServerTransportStatusType = "waitingForNextAttempt"
+	nominal               APMServerTransportStatusType = "nominal"
+)
+
+var apmServerTransportStatus APMServerTransportStatusType
+var apmServerReconnectionCount int
+var apmServerBackoffTimer *time.Timer
+
+func InitApmServerTransportStatus() {
+	apmServerTransportStatus = nominal
+}
 
 // todo: can this be a streaming or streaming style call that keeps the
 //       connection open across invocations?
@@ -75,6 +94,7 @@ func PostToApmServer(client *http.Client, agentData AgentData, config *extension
 
 	resp, err := client.Do(req)
 	if err != nil {
+		enterBackoff()
 		return fmt.Errorf("failed to post to APM server: %v", err)
 	}
 
@@ -82,10 +102,54 @@ func PostToApmServer(client *http.Client, agentData AgentData, config *extension
 	defer resp.Body.Close()
 	body, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
+		enterBackoff()
 		return fmt.Errorf("failed to read the response body after posting to the APM server")
 	}
 
 	log.Printf("APM server response body: %v\n", string(body))
 	log.Printf("APM server response status code: %v\n", resp.StatusCode)
+	apmServerTransportStatus = nominal
 	return nil
+}
+
+func EnqueueAPMData(agentDataChannel chan AgentData, agentData AgentData) {
+	select {
+	case agentDataChannel <- agentData:
+		log.Println("Adding agent data to buffer to be sent to apm server")
+	default:
+		log.Println("Channel full: dropping event")
+	}
+}
+
+func CheckTransportStatus() bool {
+	return apmServerTransportStatus != failure
+}
+
+func enterBackoff() {
+	if apmServerTransportStatus == nominal {
+		apmServerReconnectionCount = 0
+	} else {
+		apmServerReconnectionCount++
+	}
+	apmServerTransportStatus = failure
+	apmServerBackoffTimer = time.NewTimer(computeGracePeriod())
+	go waitForGracePeriod()
+}
+
+func waitForGracePeriod() {
+	<-apmServerBackoffTimer.C
+	apmServerTransportStatus = waitingForNextAttempt
+}
+
+// ComputeGracePeriod https://github.com/elastic/apm/blob/main/specs/agents/transport.md#transport-errors
+func computeGracePeriod() time.Duration {
+	var thresholdReconnectionCount float64
+	if apmServerReconnectionCount <= 6 {
+		thresholdReconnectionCount = 6
+	} else {
+		thresholdReconnectionCount = 6
+	}
+	gracePeriodWithoutJitter := math.Pow(thresholdReconnectionCount, 2)
+	jitter := rand.Float64()/5 - 0.1
+	return time.Duration(gracePeriodWithoutJitter+jitter*gracePeriodWithoutJitter) * time.Second
 }
--- a/apm-lambda-extension/main.go
+++ b/apm-lambda-extension/main.go
@@ -137,9 +137,15 @@ func main() {
 						log.Println("funcDone signal received, not processing any more agent data")
 						return
 					case agentData := <-agentDataChannel:
-						err := extension.PostToApmServer(client, agentData, config)
-						if err != nil {
-							log.Printf("Error sending to APM server, skipping: %v", err)
+						if extension.CheckTransportStatus() {
+							err := extension.PostToApmServer(client, agentData, config)
+							if err != nil {
+								log.Printf("Error sending to APM server, queueing data  : %v", err)
+								extension.EnqueueAPMData(agentDataChannel, agentData)
+							}
+						} else {
+							log.Println("Waiting for backoff grace period")
+							extension.EnqueueAPMData(agentDataChannel, agentData)
 						}
 					}
 				}
@@ -191,6 +197,11 @@ func main() {
 			backgroundDataSendWg.Wait()
 			if config.SendStrategy == extension.SyncFlush {
 				// Flush APM data now that the function invocation has completed
+				for {
+					if extension.CheckTransportStatus() {
+						break
+					}
+				}
 				extension.FlushAPMData(client, agentDataChannel, config)
 			}
 		}
--- a/apm-lambda-extension/extension/route_handlers.go
+++ b/apm-lambda-extension/extension/route_handlers.go
@@ -100,12 +100,7 @@ func handleIntakeV2Events(agentDataChan chan AgentData) func(w http.ResponseWrit
 				ContentEncoding: r.Header.Get("Content-Encoding"),
 			}
 
-			select {
-			case agentDataChan <- agentData:
-				log.Println("Adding agent data to buffer to be sent to apm server")
-			default:
-				log.Println("Channel full: dropping event")
-			}
+			EnqueueAPMData(agentDataChan, agentData)
 		}
 
 		if len(r.URL.Query()["flushed"]) > 0 && r.URL.Query()["flushed"][0] == "true" {


apm-lambda-extension/main.go
```diff
@@ -191,6 +197,11 @@ func main() {
 			backgroundDataSendWg.Wait()
 			if config.SendStrategy == extension.SyncFlush {
 				// Flush APM data now that the function invocation has completed
+				for {
+					if extension.CheckTransportStatus() {
+						break
+					}
+				}
 				extension.FlushAPMData(client, agentDataChannel, config)
```
felixbarny, line 205: Looks like this would block the Lambda runtime until the transport status is healthy again. If APM Server is down for a long period of time, this increased the billed time for all Lambda invocations and avoids the runtimes from being re-used, leading to lots of cold starts.
It also busy-spins, saturating a CPU core.

Instead, can we just flush like so:

```suggestion
				extension.FlushAPMData(client, agentDataChannel, config)
```

And exit early in

https://github.com/elastic/apm-aws-lambda/blob/8e471917dcb9e05b473243f99dcadd01a89750ea/apm-lambda-extension/extension/process_events.go#L32-L34

```go
if !IsTransportStatusHealthy() {
	return
}
```

apm-lambda-extension/extension/apm_server.go
```diff
@@ -75,17 +94,62 @@ func PostToApmServer(client *http.Client, agentData AgentData, config *extension
...
+
+// ComputeGracePeriod https://github.com/elastic/apm/blob/main/specs/agents/transport.md#transport-errors
+func computeGracePeriod() time.Duration {
+	var thresholdReconnectionCount float64
+	if apmServerReconnectionCount <= 6 {
+		thresholdReconnectionCount = 6
+	} else {
+		thresholdReconnectionCount = 6
+	}
+	gracePeriodWithoutJitter := math.Pow(thresholdReconnectionCount, 2)
```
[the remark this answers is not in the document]
jlvoiseux: Thanks for the suggestion ! When I got introduced to go, one of the first things that I learnt was that [there was no](https://stackoverflow.com/questions/27516387/what-is-the-correct-way-to-find-the-min-between-two-integers-in-go) `math.Min` function for Integers. There is one for `float64` however, and that is what we are working with here. Generics support was also introduced in Go 1.18, so using `math.Min` is definitely the way to go.

apm-lambda-extension/extension/apm_server.go
```diff
@@ -75,17 +94,62 @@ func PostToApmServer(client *http.Client, agentData AgentData, config *extension
...
+func EnqueueAPMData(agentDataChannel chan AgentData, agentData AgentData) {
+	select {
+	case agentDataChannel <- agentData:
+		log.Println("Adding agent data to buffer to be sent to apm server")
+	default:
+		log.Println("Channel full: dropping event")
+	}
+}
+
+func CheckTransportStatus() bool {
```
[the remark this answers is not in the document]
jlvoiseux: Solved in [aae7c76](https://github.com/elastic/apm-aws-lambda/pull/148/commits/aae7c76ee2ce7e4cfb4e64c380bca3b3bcf94f23)

apm-lambda-extension/main.go
```diff
@@ -137,9 +137,15 @@ func main() {
...
 						return
 					case agentData := <-agentDataChannel:
-						err := extension.PostToApmServer(client, agentData, config)
-						if err != nil {
-							log.Printf("Error sending to APM server, skipping: %v", err)
+						if extension.CheckTransportStatus() {
+							err := extension.PostToApmServer(client, agentData, config)
+							if err != nil {
+								log.Printf("Error sending to APM server, queueing data  : %v", err)
+								extension.EnqueueAPMData(agentDataChannel, agentData)
```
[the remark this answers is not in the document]
jlvoiseux: I added something similar in [aae7c76](https://github.com/elastic/apm-aws-lambda/pull/148/commits/aae7c76ee2ce7e4cfb4e64c380bca3b3bcf94f23), but I still need to figure out a way to Enqueue the data that does not result in the goroutine running perpetually.

apm-lambda-extension/extension/apm_server.go
```diff
@@ -24,14 +24,32 @@ import (
...
 }}
 
+type APMServerTransportStatusType string
+
+const (
+	failure APMServerTransportStatusType = "failing"
+	healthy APMServerTransportStatusType = "healthy"
+)
+
+var apmServerTransportStatus APMServerTransportStatusType
```
felixbarny, line 45: As it can only have two states, what about just using a bool?

apm-lambda-extension/extension/apm_server.go
```diff
@@ -75,17 +93,54 @@ func PostToApmServer(client *http.Client, agentData AgentData, config *extension
...
+}
+
+func enterBackoff() {
+	if apmServerTransportStatus == healthy {
+		apmServerReconnectionCount = 0
+	} else {
+		apmServerReconnectionCount++
+	}
+	apmServerTransportStatus = failure
+}
```
felixbarny, line 139:
```suggestion
func WaitForGracePeriod() {
	if apmServerBackoffTimer == nil {
		return
	}
	<-apmServerBackoffTimer.C
}

func enterBackoff() {
	if apmServerTransportStatus == healthy {
		apmServerReconnectionCount = 0
	} else {
		apmServerReconnectionCount++
	}
	apmServerTransportStatus = failure
	apmServerBackoffTimer = time.NewTimer(computeGracePeriod())
	go func() {
		WaitForGracePeriod()
		apmServerTransportStatus = healthy
	}()
}
```

apm-lambda-extension/extension/process_events.go
```diff
@@ -34,6 +34,9 @@ func FlushAPMData(client *http.Client, dataChannel chan AgentData, config *exten
 	for {
 		select {
 		case agentData := <-dataChannel:
+			if !IsTransportStatusHealthy() {
+				return
+			}
```
felixbarny, line 39: If we know at the beginning of the method that the status is not healthy, we don't have to remove any agent data from the channel
```suggestion
	if !IsTransportStatusHealthy() {
		return
	}
	for {
		select {
		case agentData := <-dataChannel:
```

apm-lambda-extension/main.go
```diff
@@ -137,9 +137,14 @@ func main() {
 						log.Println("funcDone signal received, not processing any more agent data")
 						return
 					case agentData := <-agentDataChannel:
+						if !extension.IsTransportStatusHealthy() {
+							extension.WaitForGracePeriod()
```
felixbarny, line 141: I just realized that blocking here might not be a good idea as this leads to the `backgroundDataSendWg.Wait()` call at the end to block, too.
Maybe just `return` from the goroutine and put the agent data back to the channel?

felixbarny: Another safety measure would be to do a 
```go
if !extension.IsTransportStatusHealthy() {
    return
}
```

right before the `for` loop

apm-lambda-extension/main.go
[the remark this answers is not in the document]
jlvoiseux: Fixed in ad48058

apm-lambda-extension/extension/apm_server.go
jlvoiseux: Solved in ad48058

apm-lambda-extension/extension/apm_server.go
```diff
@@ -24,14 +24,32 @@ import (
...
+var apmServerReconnectionCount int
+var apmServerBackoffTimer *time.Timer
+
+func InitApmServerTransportStatus() {
+	apmServerTransportStatus = healthy
+}
+
 // todo: can this be a streaming or streaming style call that keeps the
 //       connection open across invocations?
 func PostToApmServer(client *http.Client, agentData AgentData, config *extensionConfig) error {
```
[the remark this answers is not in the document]
jlvoiseux: Added in ad48058

apm-lambda-extension/extension/apm_server.go
jlvoiseux: Modified in ad48058

apm-lambda-extension/extension/apm_server.go
```diff
@@ -75,17 +91,61 @@ func PostToApmServer(client *http.Client, agentData AgentData, config *extension
...
+	select {
+	case agentDataChannel <- agentData:
+		log.Println("Adding agent data to buffer to be sent to apm server")
+	default:
+		log.Println("Channel full: dropping event")
+	}
+}
+
+func IsTransportStatusHealthy() bool {
+	return apmServerTransportFailing != true
```
felixbarny, line 122:
```suggestion
	return !apmServerTransportFailing
```

apm-lambda-extension/extension/apm_server.go
```diff
@@ -20,21 +20,37 @@ package extension
...
 	return &bytes.Buffer{}
 }}
 
+var apmServerTransportFailing bool
+var apmServerReconnectionCount int
+var apmServerBackoffTimer *time.Timer
+
+func InitApmServerTransportStatus() {
+	apmServerTransportFailing = false
+}
```
felixbarny, line 45: Not sure what's idiomatic in Go, but there are `init()` functions that are executed automatically: https://go.dev/ref/spec#Package_initialization

apm-lambda-extension/extension/apm_server.go
```diff
@@ -75,17 +91,61 @@ func PostToApmServer(client *http.Client, agentData AgentData, config *extension
...
+func enterBackoff() {
+	if !apmServerTransportFailing {
+		apmServerReconnectionCount = 0
+	} else {
+		apmServerReconnectionCount++
+	}
+	apmServerTransportFailing = true
+	apmServerBackoffTimer = time.NewTimer(computeGracePeriod())
+	go func() {
+		WaitForGracePeriod()
```
felixbarny, line 141: Maybe just sleep here instead of using a timer? `WaitForGracePeriod` isn't called outside of this file anyway.

apm-lambda-extension/main.go
```diff
@@ -131,15 +132,24 @@ func main() {
...
 				for {
 					select {
 					case <-funcDone:
 						log.Println("funcDone signal received, not processing any more agent data")
 						return
 					case agentData := <-agentDataChannel:
+						if !extension.IsTransportStatusHealthy() {
+							extension.EnqueueAPMData(agentDataChannel, agentData)
+							return
+						}
```
felixbarny, line 147: Seems redundant as if the transport status is not healthy, an error gets returned and the agent data is put back to the channel, too.

apm-lambda-extension/extension/process_events.go
jlvoiseux: Solved in ad48058

apm-lambda-extension/extension/apm_server.go
jlvoiseux: After writing some tests, I realised that an intermediate `waitingForNextAttempt` status is required, in order to avoid a lockdown situation where we do not differentiate between the situation where we enter the backoff from a healthy situation, and the situation where we enter the backoff from an already failing situation.

The `ApmServerTransportStatusType` is re-introduced in the next commits

apm-lambda-extension/main.go
jlvoiseux: Solved in 73a25ff

apm-lambda-extension/extension/apm_server.go
jlvoiseux: That is the most idiomatic solution, but I ended up needing an explicit init function in order to use unit tests (`init()` is only used once upon package initialization, which breaks unit tests when those are run sequentially)

felixbarny (approved):

apm-lambda-extension/main.go
```diff
@@ -192,7 +191,6 @@ func main() {
 			close(funcDone)
 			backgroundDataSendWg.Wait()
 			if config.SendStrategy == extension.SyncFlush {
-				// Flush APM data now that the function invocation has completed
 				extension.FlushAPMData(client, agentDataChannel, config)
 			}
```
felixbarny, line 195: Probably out of scope for this PR, but can we just move this to the `case <-funcDone:` case of the background data sending go routine?
The benefit is that the APM Server client is then only executed from a single thread and we don't have to add locks or other mechanisms to ensure visibility.

We might even want to move the background goroutine to `apm_server.go`.

In `main.go`, we would then just be calling `backgroundDataSendWg := extension.startBackgroundSending(config.SendStrategy)`.

`main.go` would just enqueue events to the `agentData` channel and `apm_server.go` is responsible for consuming from that channel. This feels like a better separation of producer/consumer concerns.

apm-lambda-extension/extension/apm_server.go
```diff
@@ -74,17 +98,57 @@ func PostToApmServer(client *http.Client, agentData AgentData, config *extension
...
+
+func enterBackoff() {
+	if apmServerTransportStatus == healthy {
+		apmServerReconnectionCount = 0
+	} else {
+		apmServerReconnectionCount++
+	}
+	apmServerTransportStatus = failing
+	go func() {
+		WaitForGracePeriod()
```
jlvoiseux, line 144: Add context to enable closure upon end of `main` for loop

apm-lambda-extension/extension/apm_server_test.go
```diff
@@ -98,9 +93,200 @@ func TestPostToApmServerDataNotCompressed(t *testing.T) {
...
+	apmServerReconnectionCount = 6
+	val6 := computeGracePeriod().Seconds()
+	assert.InDelta(t, val6, float64(36), 0.1*36)
+
+	apmServerReconnectionCount = 7
+	val7 := computeGracePeriod().Seconds()
+	assert.InDelta(t, val7, float64(36), 0.1*36)
+}
+
+func TestEnterBackoffFromNominal(t *testing.T) {
```
jlvoiseux, line 130: Replace `Nominal` by `Healthy`

apm-lambda-extension/extension/logger.go
```diff
@@ -7,7 +7,7 @@ import (
 
 var Log *logrus.Entry
 
-func InitLogger() {
+func init() {
 
 	if Log == nil {
```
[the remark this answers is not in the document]
jlvoiseux: Nice catch, thanks ! I will address it in the upcoming commits.

apm-lambda-extension/main.go
```diff
@@ -192,7 +191,6 @@ func main() {
 			close(funcDone)
 			backgroundDataSendWg.Wait()
 			if config.SendStrategy == extension.SyncFlush {
-				// Flush APM data now that the function invocation has completed
 				extension.FlushAPMData(client, agentDataChannel, config)
```
jlvoiseux, line 194: Do not remove this comment

apm-lambda-extension/main_test.go
```diff
@@ -324,6 +326,30 @@ func TestAPMServerHangs(t *testing.T) {
...
+	defer apmServer.Close()
+
+	eventsChain := []MockEvent{
+		{Type: InvokeStandard, APMServerBehavior: Hangs, ExecutionDuration: 1, Timeout: 5},
+		{Type: InvokeStandard, APMServerBehavior: TimelyResponse, ExecutionDuration: 1, Timeout: 5},
+	}
+	eventQueueGenerator(eventsChain, eventsChannel)
+	go func() {
+		time.Sleep(5 * time.Second)
+		hangChan <- struct{}{}
```
estolfo, line 346: I don't follow why we need to send twice on the hangs channel. Can you explain? Also, why is there a sleep for 5 seconds and why did you change the hangs channel to be a buffered channel of size 1?

apm-lambda-extension/extension/apm_server_test.go
jlvoiseux: Solved in 66ec239

apm-lambda-extension/extension/logger.go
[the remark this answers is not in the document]
jlvoiseux: Solved in 66ec239

apm-lambda-extension/main_test.go
jlvoiseux: Following your comment, I changed the test slightly : the channel size is now unspecified. The test itself works as follows : 

- The extension receives a first event, which contains a `Hangs` signal to make the APM server hang. The forwarding request to the APM times out (after 3 seconds, as per `ELASTIC_APM_DATA_FORWARDER_TIMEOUT_SECONDS`).

- Following the request timeout, the extension goes into backoff mode. The hanging request goes back into the channel (it is now at the front of the channel, since no other element is there). During the current invocation, no other data is sent to the APM server, as the transport status is unhealthy.

- The extension receives a second event, which is supposed to be a standard, `TimelyResponse` one. The main success condition of the test is if we manage to send this event successfully to the APM server.

- In the meantime, the grace period is over, and the extension can send data again. The APM data channel now contains two chunks : a `Hangs` and a `TimelyResponse`, in that order.

- The extension now tries again to send the `Hangs` signal to the APM server. The APM server is still hanging at that point (from the previous `Hangs` signal), and does not respond immediately.

- 5 seconds after `main()` started, the `time.Sleep()` is over, and we unblock the APM server twice : once for the previous `Hangs` signal, and once to handle the `Hangs` that we sent during the second invocation due to the backoff-enqueuing mechanism.

- The transport status is now healthy, the `TimelyResponse` signal is sent to the APM server, and the test is successful.

The reasoning behind the 5 seconds is that the main thing that dimensions the test is the default value of `ELASTIC_APM_DATA_FORWARDER_TIMEOUT_SECONDS` (3 seconds). We need to unblock the APM server at some point during the second invocation, and thus not earlier than 3 seconds.

apm-lambda-extension/extension/apm_server.go
```diff
@@ -75,17 +91,61 @@ func PostToApmServer(client *http.Client, agentData AgentData, config *extension
...
+	if !apmServerTransportFailing {
+		apmServerReconnectionCount = 0
+	} else {
+		apmServerReconnectionCount++
+	}
+	apmServerTransportFailing = true
+	apmServerBackoffTimer = time.NewTimer(computeGracePeriod())
+	go func() {
+		WaitForGracePeriod()
+		apmServerTransportFailing = false
```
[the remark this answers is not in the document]
jlvoiseux: After having a look at the current implementation, I do not think we need those anymore, as I could not figure out a scenario where `enterBackoff()` can be accessed by concurrent processes. What do you think ? From a practical perspective, Go has `Mutexes` that make it very easy to implement locks.

apm-lambda-extension/main_test.go
jlvoiseux: Make `HangLock` more explicit

jlvoiseux: Shorten forwarding timeout and put a shorter sleep time (0.5-1 sec)

apm-lambda-extension/extension/apm_server.go
```diff
@@ -72,19 +100,75 @@ func PostToApmServer(client *http.Client, agentData AgentData, config *extension
...
+func EnqueueAPMData(agentDataChannel chan AgentData, agentData AgentData) {
+	select {
+	case agentDataChannel <- agentData:
+		Log.Debug("Adding agent data to buffer to be sent to apm server")
+	default:
+		Log.Warn("Channel full: dropping a subset of agent request data")
+	}
+}
+
+func IsTransportStatusHealthy() bool {
```
estolfo, line 134: Maybe this should be called `IsTransportStatusHealthyOrPending` because I got confused when looking at the tests as this would return true if the status is `pending`

apm-lambda-extension/extension/apm_server_test.go
```diff
@@ -94,13 +90,204 @@ func TestPostToApmServerDataNotCompressed(t *testing.T) {
...
+	agentData := AgentData{Data: data, ContentEncoding: "gzip"}
+
+	// Create apm server and handler
+	apmServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		bytes, _ := ioutil.ReadAll(r.Body)
+		assert.Equal(t, string(data), string(bytes))
+		assert.Equal(t, "gzip", r.Header.Get("Content-Encoding"))
+		w.Write([]byte(`{"foo": "bar"}`))
+	}))
+	apmServer.Close()
```
estolfo, line 153: I would put a comment here saying "Close apm server before sending request" just to draw attention to what you are doing. I missed that the important difference between this test and others was that there wasn't a `defer`.

apm-lambda-extension/extension/apm_server_test.go
```diff
@@ -94,13 +90,204 @@ func TestPostToApmServerDataNotCompressed(t *testing.T) {
...
```
estolfo, line 193: Same here with calling out that you explicitly close the apm server

apm-lambda-extension/extension/apm_server_test.go
```diff
@@ -94,13 +90,204 @@ func TestPostToApmServerDataNotCompressed(t *testing.T) {
...
```
estolfo, line 278: and call out that you're closing the server here

estolfo (approved):
Overall this looks good and I'm approving but I have just a few minor comments.

apm-lambda-extension/extension/apm_server.go
```diff
@@ -72,19 +100,73 @@ func PostToApmServer(client *http.Client, agentData AgentData, config *extension
...
 	Log.Debugf("APM server response status code: %v", resp.StatusCode)
 	return nil
 }
+
+func EnqueueAPMData(agentDataChannel chan AgentData, agentData AgentData) {
+	select {
+	case agentDataChannel <- agentData:
+		Log.Debug("Adding agent data to buffer to be sent to apm server")
+	default:
+		Log.Warn("Channel full: dropping a subset of agent request data")
```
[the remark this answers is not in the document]
jlvoiseux: Solved in 240fa40

apm-lambda-extension/extension/apm_server.go
jlvoiseux: Solved in 240fa40

apm-lambda-extension/extension/apm_server_test.go
jlvoiseux: Solved in 240fa40

apm-lambda-extension/extension/apm_server_test.go
jlvoiseux: Solved in 240fa40

apm-lambda-extension/extension/apm_server_test.go
jlvoiseux: Solved in 240fa40

apm-lambda-extension/main_test.go
jlvoiseux: Comments addressed in c6663f7

apm-lambda-extension/extension/apm_server.go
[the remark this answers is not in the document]
jlvoiseux: Added a comment in c6663f7 to make sure that we keep the possibility of using mutexes in mind, should we start using `PostToAPMServer()` concurrently.

apm-lambda-extension/extension/apm_server.go
```diff
@@ -72,19 +106,79 @@ func PostToApmServer(client *http.Client, agentData AgentData, config *extension
...
+	apmServerTransportState.Status = Failing
+	Log.Debug("Transport status set to failing")
+	apmServerTransportState.GracePeriodTimer = time.NewTimer(computeGracePeriod())
+	go func() {
+		apmServerTransportState.Lock()
+		defer apmServerTransportState.Unlock()
+		WaitForGracePeriod(ctx)
+		apmServerTransportState.Status = Pending
+		Log.Debug("Transport status set to pending")
+	}()
```
jlvoiseux, line 176: Added locks in the embedded goroutine. I believe this makes sense from a design perspective (lock the transport state in `Failing` while we wait for the grace period timer to be elapsed). Such locks are also required for the tests in `apm_server_test.go` to pass : see dedicated comment.

apm-lambda-extension/extension/apm_server_test.go
```diff
@@ -94,13 +90,204 @@ func TestPostToApmServerDataNotCompressed(t *testing.T) {
...
+	}
+
+	PostToApmServer(apmServer.Client(), agentData, &config, context.Background())
+	// No way to know for sure if failing or pending (0 sec grace period)
+	assert.True(t, apmServerTransportState.Status != Healthy)
+	assert.Equal(t, apmServerTransportState.ReconnectionCount, 0)
+}
+
+func TestEnterBackoffFromFailing(t *testing.T) {
+	SetApmServerTransportStatus(Healthy, 0)
```
jlvoiseux, line 166: In previous iterations, the tests were failing because the grace period goroutine initialized during the previous test was racing against the `SetApmServerTransportStatus(Healthy, 0)` : we would call the function and set the status of the APM server transport to `Healthy`, only for the goroutine to complete its execution and to change the status of the APM server transport to `Pending`, which is not expected at the start of this test.

apm-lambda-extension/extension/apm_server_test.go
```diff
@@ -94,13 +90,204 @@ func TestPostToApmServerDataNotCompressed(t *testing.T) {
...
+	agentData := AgentData{Data: data, ContentEncoding: "gzip"}
+
+	// Create apm server and handler
+	apmServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		bytes, _ := ioutil.ReadAll(r.Body)
+		assert.Equal(t, string(data), string(bytes))
+		assert.Equal(t, "gzip", r.Header.Get("Content-Encoding"))
+		w.Write([]byte(`{"foo": "bar"}`))
+	}))
+	apmServer.Close() // Close the APM server early so that POST requests fail and that backoff is enabled
```
estolfo, line 195: I'd put the comment above the line where the server is closed

apm-lambda-extension/extension/apm_server_test.go
```diff
@@ -94,13 +90,204 @@ func TestPostToApmServerDataNotCompressed(t *testing.T) {
...
```
estolfo, line 278: I would put the comment above the line

estolfo (changes requested):
Just one minor request

apm-lambda-extension/extension/apm_server_test.go
jlvoiseux: Fixed in 384b913

apm-lambda-extension/extension/apm_server_test.go
jlvoiseux: Fixed in 384b913

apm-lambda-extension/extension/apm_server.go
```diff
@@ -72,19 +101,70 @@ func PostToApmServer(client *http.Client, agentData AgentData, config *extension
...
+
+func IsTransportStatusHealthyOrPending() bool {
+	return ApmServerTransportState.Status != Failing
+}
+
+func SetApmServerTransportState(status ApmServerTransportStatusType, ctx context.Context) {
+	ApmServerTransportState.Lock()
+	ApmServerTransportState.Status = status
+	Log.Debugf("APM Server Transport status set to %s", status)
+	switch status {
```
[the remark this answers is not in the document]
jlvoiseux: Implemented in 384b913

estolfo (approved):

PR #165, opened 2022-03-30
Force GO111MODULE=on in the makefile

### LINKED ISSUE #163: `make check-licenses` fails
```
% make check-licenses
go install github.com/elastic/go-licenser@v0.4.0
go run github.com/elastic/go-licenser@v0.4.0 -d .
package github.com/elastic/go-licenser@v0.4.0: can only use path@version syntax with 'go get' and 'go install' in module-aware mode
make: *** [check-licenses] Error 1
```

My golang ignorance means I don't immediately know what to do about "module-aware mode".

I'm using go 1.16:

```
% go version
go version go1.16.3 darwin/amd64
```

which I think is acceptable given `go 1.14` in "go.mod".  What is the min golang version for the extension?

jlvoiseux:
This PR slightly tweaks the `Makefile` to force `GO111MODULE=on`. Should resolve #163.

### DIFF
--- a/apm-lambda-extension/Makefile
+++ b/apm-lambda-extension/Makefile
@@ -2,6 +2,8 @@ SHELL = /bin/bash -eo pipefail
 
 AWS_FOLDER = .aws
 
+GO111MODULE=on
+
 ifndef GOARCH
 	GOARCH=amd64
 endif


trentm:
We decided to drop this in favour of bumping to 1.17 to fix the issue. See comment in https://github.com/elastic/apm-aws-lambda/issues/163

PR #174, opened 2022-04-01
Make logging levels compliant to APM spec

### LINKED ISSUE #170: support the spec'd values for ELASTIC_APM_LOG_LEVEL
The extension docs say these log levels are supported for `ELASTIC_APM_LOG_LEVEL`:

> Supported values are trace, debug, info, warning, error, fatal and panic.

It would be good to conform to the APM agents spec for supported values of this envvar:
https://github.com/elastic/apm/blob/main/specs/agents/logging.md#log_level-configuration

I think the extension could re-use this ParseLogLevel from the Go APM agent for that:
https://github.com/elastic/apm-agent-go/blob/main/internal/apmlog/logger.go#L152-L173

jlvoiseux:
### Motivation / Summary
This PR aims to resolve #170. The spec'd log level are now mapped on the default Logrus levels. This PR is still in draft, as I need to figure out a way to replace the `fatal` log level by `critical` in the output logs. 

Right now `ELASTIC_APM_LOG_LEVEL=critical` is supported but will map to `fatal`, which will result in outputs such as the one shown below:
```
{"@timestamp":"2022-04-01T12:22:43.425+0200","ecs.version":"1.6.0","event.dataset":"apm-lambda-extension","log.level":"fatal","message":"Could not wait for the execution of make : exit status 2"}
```
That is not ideal in terms of UX: a user who sets the log level to critical might want to see `"log.level":"critical"` for the relevant logs. This PR is in draft while I figure out a way to implement the aforementioned behavior.


apm-lambda-extension/extension/logger.go
```diff
@@ -31,3 +35,29 @@ func init() {
...
+		return logrus.InfoLevel, nil
+	case "warn", "warning":
+		// "warn" exists for backwards compatibility;
+		// "warning" is the canonical level name.
+		return logrus.WarnLevel, nil
+	case "error":
+		return logrus.ErrorLevel, nil
+	case "fatal", "critical":
+		return logrus.FatalLevel, nil
+	case "off":
```
jlvoiseux, line 58: Add `panic` again to support default logrus functionality

felixbarny: I'd suggest only using the log levels that are supported from the agent logging spec. This should give enough flexibility and allows `ELASTIC_APM_LOG_LEVEL` to be picked up by both the agent and the extension.

jlvoiseux: Ok, thanks for the suggestion, let us stick to the agent logging spec then. I was thinking that we could easily expose Logrus functionality aside from the spec-defined levels, but it is not of great use if `panic` is not supported by the agent...

jlvoiseux:
### Custom Zap-based Level Logger
The technical possibilities to modify the `log.level` field value without defining our own level system being inexistent, I decided to implement a new logger based on zap (the other Go logger for which an ECS formatter is available). Zap has the following advantages : 
- It is actively maintained, whereas Logrus is in maintenance mode
- It is faster than Logrus
- It made the implementation of this PR straightforward by offering a large set of configuration options

### Changes
- The logger is now wrapped in a struct containing its level and its config
- Functions corresponding to the APM Log Levels (`Trace()`, `Debug()`, `Info()`, `Warn()`, `Error()`, `Critical()`) are mapped with Zap levels as shown below:
```
Zap Levels              APM Levels
DebugLevel              trace, debug
InfoLevel               info
WarnLevel               warn(ing)
ErrorLevel              error
DPanicLevel             /
PanicLevel              /
FatalLevel              critical
```

- An `Off` level was implemented.

jlvoiseux:
@axw Thanks for taking a look! I believe that the reasoning behind having the same values for `ELASTIC_APM_LOG_LEVEL` across the extension and the agent is that the env variable actually affects both (ie with a Node Lambda function, we will see `debug` logs from the Node agent if `ELASTIC_APM_LOG_LEVEL` is set) hence the incentive to follow the spec.

However I agree with you, the PR could be simplified by a lot if we decide that this incentive is not worth having a 1:1 mapping.

axw:
@jlvoiseux `ELASTIC_APM_LOG_LEVEL` should accept the same values regardless of agent. What I meant is that I don't think we need the additional `extension.LevelLogger` type, or to explicitly set `log.level`.

If we use `zap` directly, and configure it with `ecszap` and the level parsed from `ELASTIC_APM_LOG_LEVEL` and translated to Zap's levels, then `log.level` will be added to log records. They won't match what is defined in `ELASTIC_APM_LOG_LEVEL`; they will use the native Zap levels.

jlvoiseux:
@axw I have implemented your recommendations. Zap is now used natively. The mapping is now as follows:
```
Zap Levels              APM Levels
DebugLevel              trace, debug
InfoLevel               info
WarnLevel               warn(ing)
ErrorLevel              error
DPanicLevel             /
PanicLevel              critical
FatalLevel              off
```
Mapping `off` to `FatalLevel` simplifies the code a lot, but that is an arbitrary choice. What do you think?

docs/monitoring-aws-lambda.asciidoc
```diff
@@ -76,4 +76,4 @@ the next request until the extension has flushed all the data. This has a negati
 though it ensures that all APM data is sent to the APM server.
 
 === `ELASTIC_APM_LOG_LEVEL`
-The logging level to be used by both the APM Agent and the Lambda Extension. Supported values are `trace`, `debug`, `info`, `warning`, `error`, `fatal` and `panic`.
+The logging level to be used by both the APM Agent and the Lambda Extension. Supported values are `trace`, `debug`, `info`, `warning`, `error`, `critical` and `off` (the latter disables logging, except for fatal errors).
```
axw, line 79: > `off` (the latter disables logging, except for fatal errors).

I think it's a bit surprising that we would log when the level is "off", even for fatal level. Can you please explain the rationale behind this?

apm-lambda-extension/extension/logger_test.go
```diff
@@ -0,0 +1,90 @@
...
+func TestLoggerSetLogLevel(t *testing.T) {
+	tempFile, err := ioutil.TempFile(os.TempDir(), "tempFileLoggerTest-")
+	require.NoError(t, err)
+	defer tempFile.Close()
+
+	Log.Level.SetLevel(zapcore.DebugLevel)
+
+	defer func() {
+		Log.Level.SetLevel(zapcore.InfoLevel)
+	}()
```
axw, line 81:
```suggestion
	defer Log.Level.SetLevel(zapcore.InfoLevel)
```

apm-lambda-extension/extension/logger_test.go
```diff
@@ -0,0 +1,90 @@
...
+	assert.Equal(t, zapcore.PanicLevel, critical)
+	assert.Equal(t, zapcore.FatalLevel, off)
+	assert.Equal(t, zapcore.InfoLevel, invalid)
+	assert.Error(t, err, "invalid log level string Inva@Lid3")
+}
+
+func TestLoggerSetLogLevel(t *testing.T) {
+	tempFile, err := ioutil.TempFile(os.TempDir(), "tempFileLoggerTest-")
+	require.NoError(t, err)
+	defer tempFile.Close()
```
axw, line 75:
```suggestion
	defer os.RemoveAll(tempFile.Name())
	defer tempFile.Close()
```

From the docs:

> It is the caller's responsibility to remove the file when no longer needed.

Alternatively you could use [`testing.T.TempDir`](https://pkg.go.dev/testing#T.TempDir), which is automatically removed at the end of the test.

apm-lambda-extension/extension/logger_test.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+func TestInitLogger(t *testing.T) {
+	assert.NotNil(t, Log)
+}
+
+func TestDefaultLogger(t *testing.T) {
+	tempFile, err := ioutil.TempFile(os.TempDir(), "tempFileLoggerTest-")
+	require.NoError(t, err)
+	defer tempFile.Close()
```
axw, line 40: The file needs to be removed too -- see next comment.

apm-lambda-extension/extension/logger.go
```diff
@@ -18,16 +18,62 @@
...
+	case "warn", "warning":
+		// "warn" exists for backwards compatibility;
+		// "warning" is the canonical level name.
+		return zapcore.WarnLevel, nil
+	case "error":
+		return zapcore.ErrorLevel, nil
+	case "critical":
+		return zapcore.PanicLevel, nil
+	case "off":
+		return zapcore.FatalLevel, nil
```
axw, line 67: I think we should map critical to fatal, and off to `zapcore.FatalLevel+1` to effectively disable all logging.

axw (commented):
Thanks for the changes, it looks a lot more straightforward to me now.

The mapping looks good except for critical and off - I think off should really mean off.

docs/monitoring-aws-lambda.asciidoc
jlvoiseux: I will implement a mechanism to remove logging when users specify `off` then (ie redirect logging output to /dev/null), as there is none built-in with Zap. 

The rationale behind the current implementation stems from:
- The lack of built-in mechanism (I wanted to strip down the implementation to a simple mapping to Zap levels)
- This [answer](https://github.com/uber-go/zap/issues/438) in the Zap repo

jlvoiseux: Implemented in 508175e

apm-lambda-extension/extension/logger_test.go
jlvoiseux: Fixed in 508175e

apm-lambda-extension/extension/logger_test.go
jlvoiseux: I ended up using `testing.T.TempDir` in 508175e

axw (approved):
LGTM, thanks for the changes!

PR #182, opened 2022-04-14
Make info request timeout configurable

jlvoiseux:
### Summary
This pull request propagates the `ELASTIC_APM_DATA_FORWARDER_TIMEOUT_SECONDS` parameter to the configuration of the reverse proxy tasked with serving APM Server info requests. A error handler is also defined to enable backoff upon transaction failure. Associated tests have also been written in `main_test`.

With this PR, backoff should be enabled after all failed attempts to communicate with the APM Server.

### Changes
- Propagate `ctx` and `config` through `handleInfoRequest`
- Set the `ResponseHeaderTimeout` of the reverse proxy to `ELASTIC_APM_DATA_FORWARDER_TIMEOUT_SECONDS` 
- Define an error handler triggered upon transport failure
- Modify `main_test` accordingly

### How to test
- Set up a slow / unresponsive APM Server
- Setup a basic Python APM-instrumented Lambda function
- Set an large timeout (30+ seconds)
- Upon cold start, the Lambda function execution should last between 3 to 5 seconds, and an error containing `Error querying version from the APM Server` should be logged.


apm-lambda-extension/extension/route_handlers.go
```diff
@@ -30,18 +32,28 @@ type AgentData struct {
...
 		if err != nil {
 			Log.Errorf("could not parse APM server URL: %v", err)
 			return
 		}
+
 		reverseProxy := httputil.NewSingleHostReverseProxy(parsedApmServerUrl)
+		reverseProxyTimeout := time.Duration(config.DataForwarderTimeoutSeconds) * time.Second
+		reverseProxy.Transport = &http.Transport{
+			ResponseHeaderTimeout: reverseProxyTimeout,
+		}
```
simitt, line 55: As mentioned offline, please do not overwrite the whole `http.Transport` here, but rather only set the `ResponseHeaderTimeout`, or even pass it as an option above when calling `httputil.NewSingleHostReverseProxy`.

simitt, line 55: As mentioned offline, please do not overwrite the whole `http.Transport` here, but rather only set the `ResponseHeaderTimeout`, or even pass it as an option above when calling `httputil.NewSingleHostReverseProxy`.

simitt (commented):
Apart from the comment about the transport, PR LGTM.

apm-lambda-extension/extension/route_handlers.go
jlvoiseux: After both reviewing the [implementation](https://cs.opensource.google/go/go/+/master:src/net/http/httputil/reverseproxy.go;drc=a10a209b23f877c80d8a5f3ebda1ce4b492ac3a9;l=144) of `NewSingleHostReverseProxy` and debugging the current implementation, it seems that the the default reverse proxy comes with a `nil` transport, which then gets [processed](https://cs.opensource.google/go/go/+/master:src/net/http/httputil/reverseproxy.go;drc=a10a209b23f877c80d8a5f3ebda1ce4b492ac3a9;l=215) as `http.DefaultTransport` by `ServeHTTP()`.

image

To alleviate the fears discussed online, I have [replaced](
2240711) the creation of a `Transport` from scratch by a modification of `http.DefaultTransport`, which is the Transport that is used anyway.

apm-lambda-extension/extension/route_handlers.go
```diff
@@ -30,19 +32,30 @@ type AgentData struct {
...
 		parsedApmServerUrl, err := url.Parse(apmServerUrl)
 		if err != nil {
 			Log.Errorf("could not parse APM server URL: %v", err)
 			return
 		}
+
 		reverseProxy := httputil.NewSingleHostReverseProxy(parsedApmServerUrl)
 
+		reverseProxyTimeout := time.Duration(config.DataForwarderTimeoutSeconds) * time.Second
+		reverseProxy.Transport = http.DefaultTransport
```
simitt, line 54: The caller of `httputil.NewSingleHostReverseProxy(parsedApmServerUrl)` should not need to have any knowledge about internas of how the `reverseProxy` is built. With this change the whole `transport` of the `reverseProxy` is still switched out. What I originally meant, was that only the `timeout` should eb customized here, or even better, be passed into the abstracting `httputil.NewSingleHostReverseProxy` method.

PR #127, opened 2022-02-18
Fetch Lambda Platform Metrics

jlvoiseux:
### Motivation / Summary
Following issue #108, the goal of this pull request is to send the following metrics to Elasticsearch through the AWS Lambda APM extension : 
- Init Duration
- Total Duration
- Billed Duration
- Memory Used
- Memory Sized

This information is typically provided by the [Lambda Platform Logs API](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-logs-api.html) as follows : 
```
REPORT RequestId: cc18960e-80e4-4160-bfdc-a5dae659993b	Duration: 162.81 ms	Billed Duration: 163 ms	Memory Size: 128 MB	Max Memory Used: 79 MB	Init Duration: 646.33 ms
```
Current demo dashboard : 
image

Memory usage is also displayed in the APM UI : 
image

### Files / Added Elements
- A new file, `process_metrics.go` has been added to the extension (as well as the corresponding test file). This file contains the `structs` required to implements the Metrics model. Most of the code comes from the Go Agent codebase, with some simplifications (most notably at the Metrics aggregation level).
- In the `logs_api` folder, the `client` and the `http_listener` have been modified to take events of type `Report` into account.
- The main function has been modified to call the `process_metrics` functions upon dedication of a `Report` event.

### Limitations / Elements to discuss
- The Platform Logs report is only made available *after* the function is over. As such, metrics are only sent if another invocation is triggered in a timespan short enough for the function to leave its warm state. As per definition, all workarounds would extend the Lambda runtime and thus incur additional billing for users.
- Metadata retrieval is currently very limited, and done minimally through a mix of hard-coded values fitting the test example and environment variables. There are several ways to populate the metadata - create a dedicated endpoint on the extension for the agents to send the information, or decompress and analyse the first received transaction.
- The newly created fields have to be added manually to the APM index pattern for reproduction.

### Checklist

- [x] Provide a first proof of concept
- [x] Devise and implement a strategy for metadata retrieval
- [ ] Confirm the names of the newly added fields
- [ ] Refine proof of concept and dashboard
- [x] Settle on a final list of metrics to collect and expand scope
- [ ] Expose Lambda specific metrics via UI

### How to test
0. Make sure that your Go Runtime is at least 1.17
1. Clone/fork the current repository and switch to the following [branch](https://github.com/jlvoiseux/apm-aws-lambda?organization=jlvoiseux&organization=jlvoiseux)
2. Build a new version of the extension and save it as a layer
[...]

PR #191, opened 2022-04-26
Restructure Packages and remove State Variables

### LINKED ISSUE #143: Investigate using a context passed to the logs API subscribe function
We might want to send a specific `context.Context` to the Subscribe function that can be used when the logs API server is started. That way, we can explicitly shutdown the server and not interrupt any running requests when the extension is shutting down.
See this discussion [here](https://github.com/elastic/apm-aws-lambda/pull/129#discussion_r816895287)

### LINKED ISSUE #151: Move the APM Data processing goroutine to apm_server.go
We can move the call to `FlushAPMData()` to the goroutine tasked with processing APM data chunks (as a follow-up to `case <- funcDone`), and then move this goroutine to `apm_server.go`.

Quoting @felixbarny in PR #148:  

> The benefit is that the APM Server client is then only executed from a single thread and we don't have to add locks or other mechanisms to ensure visibility. 

> We might even want to move the background goroutine to apm_server.go. In main.go, we would then just be calling `backgroundDataSendWg := extension.startBackgroundSending(config.SendStrategy)`.

> `main.go` would just enqueue events to the agentData channel and apm_server.go is responsible for consuming from that channel. This feels like a better separation of producer/consumer concerns.

### LINKED ISSUE #178: Replace package-level variable by main-instantiated structs
As mentioned in #169, the extension uses a lot of package-level variables, instantiated in the various files of the extension package. The most notable are:
- The logger
- The `APMServerState`
- The `AgentDataServer`
- The `AgentDone` signal

Replacing these variables by structs instantiated by `main.go` will make it easier to track the status of the extension, especially across multiple tests. The tests could create their own instances of the structs and then be run in parallel.

jlvoiseux:
### Motivation / Summary
This PR aims to resolve several tech-debt-related issues, as well as to integrate improvements discussed during past code reviews.

### Changes

- [x]  Resolve #143
- [x] Resolve #151 
- [x] Resolve #178 
- [x] ~~Move logger to the main package and replace function logging by error processing~~ Out of scope, relevant issue to be created (#192)

### How to test
- Switch to the related branch, [create and upload the extension as a layer and use it to instrument an empty Lambda function.](https://github.com/jlvoiseux/elastic-apm-lambda-dev-sample)
- Alternatively, the following [repo](https://github.com/elastic/apm-lambda-extension-testing) can be used with the corresponding release.

### DIFF
--- a/apm-lambda-extension/logsapi/subscribe.go
+++ b/apm-lambda-extension/logsapi/subscribe.go
@@ -148,8 +148,8 @@ func checkAWSSamLocal() bool {
 // It is currently used together with checkAWSSamLocal as a best effort solution to determine if
 // the extension actually runs in dev (unit tests), SAM, or in a local Lambda environment.
 func checkLambdaFunction() bool {
-	_, ok := os.LookupEnv("AWS_LAMBDA_FUNCTION_NAME")
-	if ok {
+	lambdaName, ok := os.LookupEnv("AWS_LAMBDA_FUNCTION_NAME")
+	if ok && lambdaName != "" {
 		return true
 	}
 
--- a/apm-lambda-extension/logsapi/subscribe_test.go
+++ b/apm-lambda-extension/logsapi/subscribe_test.go
@@ -44,6 +44,20 @@ func TestSubscribeWithSamLocalEnv(t *testing.T) {
 	assert.Error(t, err)
 }
 
+func TestSubscribeWithLambdaFunction(t *testing.T) {
+	if err := os.Setenv("AWS_LAMBDA_FUNCTION_NAME", "mock"); err != nil {
+		t.Fail()
+	}
+	t.Cleanup(func() {
+		if err := os.Unsetenv("AWS_LAMBDA_FUNCTION_NAME"); err != nil {
+			t.Fail()
+		}
+	})
+
+	_, err := Subscribe(context.Background(), "testID", []EventType{Platform})
+	assert.Error(t, err, "listen tcp: lookup sandbox: no such host")
+}
+
 func TestSubscribeAWSRequest(t *testing.T) {
 
 	// For subscription request


apm-lambda-extension/extension/apm_server.go
```diff
@@ -48,30 +44,70 @@ const (
...
+	transport.ctx = ctx
+	transport.Status = Healthy
+	transport.ReconnectionCount = -1
+	return &transport
+}
+
+// StartBackgroundApmDataForwarding Receive agent data as it comes in and post it to the APM server.
+// Stop checking for, and sending agent data when the function invocation
+// has completed, signaled via a channel.
+func StartBackgroundApmDataForwarding(transport *ApmServerTransport, funcDone chan struct{}, backgroundDataSendWg *sync.WaitGroup) {
```
felixbarny, line 81: Not a go expert but maybe this and other methods in this class should be receiver functions

```suggestion
func(transport *ApmServerTransport) StartBackgroundApmDataForwarding(funcDone chan struct{}, backgroundDataSendWg *sync.WaitGroup) {
```

apm-lambda-extension/extension/process_events.go
```diff
@@ -18,30 +18,30 @@
...
 	Log.Info("Received SHUTDOWN event, exiting")
 	agentDataServer.Close()
+	logsServer.Close()
 }
 
 // FlushAPMData reads all the apm data in the apm data channel and sends it to the APM server.
-func FlushAPMData(client *http.Client, dataChannel chan AgentData, config *extensionConfig, ctx context.Context) {
-	if !IsTransportStatusHealthyOrPending() {
-		Log.Debug("Flush skipped - Transport unhealthy")
+func FlushAPMData(transport *ApmServerTransport) {
```
felixbarny, line 34: This function looks very similar to `StartBackgroundApmDataForwarding`. Probably a sign that it should be moved to `apm_server.go`.

jlvoiseux: Thank you for this comment! It should definitely be - actually, I think we have a lot of room for improvement when it comes to the structure of the extension - see #192.

apm-lambda-extension/extension/apm_server.go
```diff
@@ -48,30 +44,70 @@ const (
 
 // A struct to track the state and status of sending
 // to the APM server. Used in the backoff implementation.
-type ApmServerTransportStateType struct {
+type ApmServerTransport struct {
 	sync.Mutex
+	bufferPool        sync.Pool
```
jlvoiseux, line 49: Check if all exported fields need to be exported

apm-lambda-extension/logsapi/subscribe.go
```diff
@@ -60,45 +77,52 @@ func subscribe(extensionID string, eventTypes []EventType) error {
...
+		extension.Log.Infof("Extension listening for Lambda Logs API events on %s", transport.Listener.Addr().String())
+		if err = transport.Server.Serve(transport.Listener); err != nil {
 			extension.Log.Errorf("Error upon Logs API server start : %v", err)
 		}
 	}()
+
+	go func() {
+		<-transport.ctx.Done()
+		transport.Server.Close()
+	}()
```
jlvoiseux, line 124: Should context be taken out of the transport struct?

apm-lambda-extension/main.go
```diff
@@ -125,68 +117,20 @@ func main() {
...
-								runtimeDoneSignal <- struct{}{}
-								return
-							} else {
-								extension.Log.Debug("Log API runtimeDone event request id didn't match")
-							}
-						}
-					}
-				}
-			}()
+			extension.StartBackgroundApmDataForwarding(apmServerTransport, funcDone, &backgroundDataSendWg)
```
axw, line 132: Can you please remind me why we only forward data to APM Server while the invocation is active? Why don't we start forwarding the data in the background outside the loop?

Is it to do with flushing at the end of the invocation? If that is the case, then perhaps we could find an alternative means of flushing, e.g. call StartBackgroundApmDataForwarding once before the loop, and allow it to receive requests over a channel to flush.

axw (commented):
Here's some initial feedback.

Would it be possible to break this up a little bit? Maybe even just splitting 393aeefe113b8f9617455a9a055171edf94bde44 out into a separate PR? It's difficult to provide a comprehensive review when there are unrelated changes.

apm-lambda-extension/main.go
jlvoiseux: Lambda users are billed based on the total execution time of their function - this includes extension execution time. To quote the [docs](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-extensions-api.html):

> Extensions can impact function performance. As an extension author, you have control over the performance impact of your extension. For example, if your extension performs compute-intensive operations, the function's duration increases because the extension and the function code share the same CPU resources. In addition, if your extension performs extensive operations after the function invocation completes, the function duration increases because the Invoke phase continues until all extensions signal that they are completed.

However, your suggested implementation makes sense - if the Lambda environment is able to freeze the main for loop, why would we need to tie the data-forwarding goroutine to a single event?

However, after running a few tests, while getting the goroutines outside of the main for loop allows for successful APM data chunks forwarding, the goroutine is still running by the time we reach the end of the current invocation and will prevent the main thread from being frozen at the end of the active invocation, leading to the Lambda function execution reaching the defined timeout.

apm-lambda-extension/logsapi/subscribe.go
jlvoiseux: Fixed in 176e264

apm-lambda-extension/main.go
```diff
@@ -71,34 +70,27 @@ func main() {
...
-	logsChannel := make(chan logsapi.LogEvent)
+	logsTransport := logsapi.InitLogsTransport(ctx)
 
 	// Use a wait group to ensure the background go routine sending to the APM server
 	// completes before signaling that the extension is ready for the next invocation.
 	var backgroundDataSendWg sync.WaitGroup
 
 	if err := logsapi.Subscribe(
-		ctx,
+		logsTransport,
```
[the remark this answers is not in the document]
jlvoiseux: Great recommendation! Implemented in 7c24276

A small modification I made was making `Subscribe` return a `*LogsTransport`. This is used at two places in `main.go`:
- As an argument to `ProcessShutdown`. I have removed it, as `ProcessShutdown` will be removed in a later commit as per a further comment.
- `LogsTransport.RuntimeDoneSignal` is used as a failsafe to make sure that we freeze the extension as soon as the current invocation is over.

jlvoiseux:
@axw Thank you for your feedback! I have rebased 393aeef - that one will be the object of a later PR.

apm-lambda-extension/main.go
```diff
@@ -125,68 +117,20 @@ func main() {
...
-								return
-							} else {
-								extension.Log.Debug("Log API runtimeDone event request id didn't match")
-							}
-						}
-					}
-				}
-			}()
+			extension.StartBackgroundApmDataForwarding(apmServerTransport, funcDone, &backgroundDataSendWg)
+			logsapi.StartBackgroundLogsProcessing(logsTransport, funcDone, event.RequestID)
```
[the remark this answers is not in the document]
jlvoiseux: Implemented in 2870ee7. `WaitRuntimeDone` will be heavily modified in #127, as metrics are obtained through the logs API.

apm-lambda-extension/main.go
```diff
@@ -197,9 +141,9 @@ func main() {
 			defer timer.Stop()
```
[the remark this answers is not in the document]
jlvoiseux: Implemented in 2870ee7. However, I had to move the Flush function out of that separate function to avoid a context deadlock after replacing `funcDone` by a context: 
- The invocation context could not be closed if the goroutine was still running due to the waitgroup
- The goroutine could not be interrupted by any other means than the invocation context being cancelled

apm-lambda-extension/main.go
```diff
@@ -125,68 +117,20 @@ func main() {
...
 			// Make a channel for signaling that the function invocation is complete
 			funcDone := make(chan struct{})
 
 			// A shutdown event indicates the execution environment is shutting down.
 			// This is usually due to inactivity.
 			if event.EventType == extension.Shutdown {
-				extension.ProcessShutdown()
+				extension.ProcessShutdown(agentDataServer, logsTransport.Server)
 				cancel()
 				return
```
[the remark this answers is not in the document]
jlvoiseux: Implemented in 2870ee7. I believe I still have to manually close the global context, unless I am missing something.

apm-lambda-extension/extension/apm_server.go
jlvoiseux: Implemented in 0855915.

apm-lambda-extension/extension/process_events.go
jlvoiseux: Implemented in 0855915

apm-lambda-extension/extension/apm_server.go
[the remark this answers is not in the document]
jlvoiseux: Replaced `funcDone` by a context in 2870ee7.
Implemented Felix's recommendations in 0855915.

apm-lambda-extension/main.go
```diff
@@ -71,35 +69,19 @@ func main() {
...
 
-	// Create a channel to buffer apm agent data
-	agentDataChannel := make(chan extension.AgentData, 100)
-
-	// Start http server to receive data from agent
-	if err = extension.StartHttpServer(ctx, agentDataChannel, config); err != nil {
+	// Init APM Server Transport struct and start http server to receive data from agent
+	apmServerTransport := extension.InitApmServerTransport(config)
+	agentDataServer, err := extension.StartHttpServer(ctx, apmServerTransport)
+	if err != nil {
```
jlvoiseux, line 75: There might be a way to remove one of these two calls, in a similar fashion to what we did with `logsapi.Subscribe` and `logsTransport`.

apm-lambda-extension/extension/apm_server.go
jlvoiseux: Fixed in 63cd01e

apm-lambda-extension/extension/route_handlers.go
```diff
@@ -31,30 +31,31 @@ type AgentData struct {
...
 
-		reverseProxyTimeout := time.Duration(config.DataForwarderTimeoutSeconds) * time.Second
+		reverseProxyTimeout := time.Duration(apmServerTransport.config.DataForwarderTimeoutSeconds) * time.Second
 		customTransport := http.DefaultTransport.(*http.Transport).Clone()
 		customTransport.ResponseHeaderTimeout = reverseProxyTimeout
 		reverseProxy.Transport = customTransport
 
+		currApmServerTransport = apmServerTransport
+		currContext = ctx
 		reverseProxy.ErrorHandler = reverseProxyErrorHandler
```
axw, line 59:
```suggestion
		reverseProxy.ErrorHandler = func(w http.ResponseWriter, r *http.Request, err error) {
			SetApmServerTransportState(ctx, apmServerTransport, Failing)
			Log.Errorf("Error querying version from the APM server: %v", err)
		}
```

Then we don't need to store globals.

axw (commented):
Thanks @jlvoiseux, looking good.

I'm a bit concerned about races between agents requesting a flush, the completion of the invocation according to runtimeDone, and the flush timer. As it is we could mistakenly think we should flush earlier than we should. I'd like to dig into the requirements a bit here to see if we can simplify things.

axw:
> I'm a bit concerned about races between agents requesting a flush, the completion of the invocation according to runtimeDone, and the flush timer. As it is we could mistakenly think we should flush earlier than we should. I'd like to dig into the requirements a bit here to see if we can simplify things.

After some offline discussions, it seems like what we have is desirable. It would be good to document some of the reasoning in the code (maybe at the select loop in processEvent) for posterity.

Still, we need to ensure signals cannot leak between invocations.

apm-lambda-extension/main.go
```diff
@@ -108,109 +90,78 @@ func main() {
...
+	durationUntilFlushDeadline := time.Until(time.Unix(flushDeadlineMs/1000, 0))
+
+	// Create a timer that expires after durationUntilFlushDeadline
+	timer := time.NewTimer(durationUntilFlushDeadline)
+	defer timer.Stop()
+
+	select {
+	case <-apmServerTransport.AgentDoneSignal:
+		extension.Log.Debug("Received agent done signal")
+	case <-logsTransport.RuntimeDoneSignal:
```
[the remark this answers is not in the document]
jlvoiseux: Fixed in 3c198fc and b60383a

apm-lambda-extension/logsapi/subscribe.go
```diff
@@ -29,9 +29,24 @@ import (
...
+	listenerHost      string
+	RuntimeDoneSignal chan struct{}
+	server            *http.Server
+}
+
+func InitLogsTransport() *LogsTransport {
+	var transport LogsTransport
+	transport.listenerHost = "localhost"
+	transport.logsChannel = make(chan LogEvent, 100)
+	transport.RuntimeDoneSignal = make(chan struct{}, 1)
```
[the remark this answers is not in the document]
jlvoiseux: Fixed in 3c198fc and b60383a

apm-lambda-extension/logsapi/subscribe.go
```diff
@@ -29,9 +29,24 @@ import (
...
-var Server *http.Server
-var Listener net.Listener
+// TestListenerAddr For e2e testing purposes
+var TestListenerAddr net.Addr
+
+type LogsTransport struct {
+	logsChannel       chan LogEvent
+	listener          net.Listener
+	listenerHost      string
+	RuntimeDoneSignal chan struct{}
```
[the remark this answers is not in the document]
jlvoiseux: Fixed in 3c198fc and b60383a

apm-lambda-extension/logsapi/subscribe.go
```diff
@@ -110,3 +138,28 @@ func checkAWSSamLocal() bool {
...
+	for {
+		select {
+		case logEvent := <-transport.logsChannel:
+			extension.Log.Debugf("Received log event %v", logEvent.Type)
+			// Check the logEvent for runtimeDone and compare the RequestID
+			// to the id that came in via the Next API
+			if logEvent.Type == RuntimeDone {
+				if logEvent.Record.RequestId == requestID {
+					extension.Log.Info("Received runtimeDone event for this function invocation")
+					transport.RuntimeDoneSignal <- struct{}{}
```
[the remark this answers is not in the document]
jlvoiseux: Fixed in 3c198fc and b60383a

apm-lambda-extension/main.go
```diff
@@ -108,109 +90,78 @@ func main() {
...
+		cancel()
+		return
+	}
+
+	// APM Data Processing
+	backgroundDataSendWg.Add(1)
+	go func() {
+		if err := apmServerTransport.ForwardApmData(invocationCtx, backgroundDataSendWg); err != nil {
+			extension.Log.Error(err)
+		}
```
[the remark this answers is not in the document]
jlvoiseux: Implemented in 2f0825d

apm-lambda-extension/logsapi/subscribe.go
```diff
@@ -29,9 +29,24 @@ import (
 	"github.com/pkg/errors"
 )
 
-var ListenerHost = "sandbox"
-var Server *http.Server
-var Listener net.Listener
+// TestListenerAddr For e2e testing purposes
```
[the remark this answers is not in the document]
jlvoiseux: TODO added in cb0f189. Related issue created (#193)

apm-lambda-extension/extension/apm_server.go
```diff
@@ -48,30 +44,85 @@ const (
...
+		case agentData := <-transport.dataChannel:
+			if err := PostToApmServer(ctx, transport, agentData); err != nil {
+				return fmt.Errorf("error sending to APM server, skipping: %v", err)
+			}
+		}
+	}
+}
+
+// FlushAPMData reads all the apm data in the apm data channel and sends it to the APM server.
+func FlushAPMData(ctx context.Context, transport *ApmServerTransport) {
```
felixbarny, line 97:
```suggestion
func (transport *ApmServerTransport) FlushAPMData(ctx context.Context) {
```

apm-lambda-extension/extension/apm_server.go
```diff
@@ -48,30 +44,85 @@ const (
...
+	}
 }
 
 // PostToApmServer takes a chunk of APM agent data and posts it to the APM server.
 //
 // The function compresses the APM agent data, if it's not already compressed.
 // It sets the APM transport status to failing upon errors, as part of the backoff
 // strategy.
-func PostToApmServer(client *http.Client, agentData AgentData, config *extensionConfig, ctx context.Context) error {
+func PostToApmServer(ctx context.Context, transport *ApmServerTransport, agentData AgentData) error {
```
felixbarny, line 122:
```suggestion
func (transport *ApmServerTransport) PostToApmServer(ctx context.Context, agentData AgentData) error {
```

apm-lambda-extension/extension/apm_server.go
```diff
@@ -101,89 +152,81 @@ func PostToApmServer(client *http.Client, agentData AgentData, config *extension
...
-
 // SetApmServerTransportState takes a state of the APM server transport and updates
 // the current state of the transport. For a change to a failing state, the grace period
 // is calculated and a go routine is started that waits for that period to complete
 // before changing the status to "pending". This would allow a subsequent send attempt
 // to the APM server.
 //
 // This function is public for use in tests.
-func SetApmServerTransportState(status ApmServerTransportStatusType, ctx context.Context) {
+func SetApmServerTransportState(ctx context.Context, transport *ApmServerTransport, status ApmServerTransportStatusType) {
```
felixbarny, line 196:
```suggestion
func (transport *ApmServerTransport) SetApmServerTransportState(ctx context.Context, status ApmServerTransportStatusType) {
```

apm-lambda-extension/extension/apm_server.go
```diff
@@ -101,89 +152,81 @@ func PostToApmServer(client *http.Client, agentData AgentData, config *extension
...
 	default:
-		Log.Errorf("Cannot set APM Server Transport status to %s", status)
+		Log.Errorf("Cannot set APM server Transport status to %s", status)
 	}
 }
 
 // ComputeGracePeriod https://github.com/elastic/apm/blob/main/specs/agents/transport.md#transport-errors
-func computeGracePeriod() time.Duration {
-	gracePeriodWithoutJitter := math.Pow(math.Min(float64(ApmServerTransportState.ReconnectionCount), 6), 2)
+func computeGracePeriod(transport *ApmServerTransport) time.Duration {
```
felixbarny, line 228:
```suggestion
func (transport *ApmServerTransport) computeGracePeriod() time.Duration {
```

apm-lambda-extension/extension/route_handlers.go
jlvoiseux: Implemented in 457e247

apm-lambda-extension/extension/route_handlers.go
```diff
@@ -31,30 +31,31 @@ type AgentData struct {
 	ContentEncoding string
 }
 
-var AgentDoneSignal chan struct{}
-var mainExtensionContext context.Context
+var currApmServerTransport *ApmServerTransport
+var currContext context.Context
```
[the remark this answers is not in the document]
jlvoiseux: Implemented in 457e247

apm-lambda-extension/extension/apm_server.go
jlvoiseux: Implemented in https://github.com/elastic/apm-aws-lambda/commit/d71ccfb98277e9d98fd558d9fd3be9cf796a6515

apm-lambda-extension/extension/apm_server.go
jlvoiseux: Implemented in https://github.com/elastic/apm-aws-lambda/commit/d71ccfb98277e9d98fd558d9fd3be9cf796a6515

apm-lambda-extension/extension/apm_server.go
jlvoiseux: Implemented in https://github.com/elastic/apm-aws-lambda/commit/d71ccfb98277e9d98fd558d9fd3be9cf796a6515

jlvoiseux:
@axw, @felixbarny, thank you for your reviews. I have implemented the related corrections last Friday.

To solve the race related to `AgentDoneSignal`, I have reproduced it by adding a test in `main_test.go`. This test simulates the case where the `AgentDoneSignal` corresponding to the invocation `i` is received after the `RuntimeDoneSignal`, thus interrupting the execution of the invocation `i+1`.

This scenario is avoided by opening and closing the `AgentDoneSignal` for each Lambda invocation. A limitation of this solution is that the edge case where the `AgentDoneSignal` of the invocation `i` is received *during* the invocation `i+1` - but the only surefire way to handle this case would be to decompress the APM Payload to check `faas.execution_id`. Moreover, this scenario happening is very unlikely, as the Agent will have been stopped and restarted by the Lambda service by the time we reach the next invocation.

axw (approved):
Thanks for the updates @jlvoiseux. Although there's still a theoretical race with AgentDoneSignal, it's at least no worse than it was before. Maybe we can eliminate it in the future.

jlvoiseux:
**Logs API Issue Fix**
The issue mentioned offline was due to me overlooking the following [documentation](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-logs-api.html) item:

> HTTP (recommended) – Lambda delivers logs to a local HTTP endpoint (http://sandbox.localdomain:${PORT}/${PATH}) as an array of records in JSON format. The $PATH parameter is optional. Note that only HTTP is supported, not HTTPS. You can choose to receive logs through PUT or POST.

Prior to the PR, [the default listener address was set as a global variable](https://github.com/elastic/apm-aws-lambda/pull/191/files#diff-a648b8c955925851657a26f9a6f1f8957315f585930586b7d8faa636e9cc77ad) to `sandbox` and then modified to accommodate our tests needs. This is no longer the case thanks to the implementation of `InitLogsTransport()`. However, the default listener address set by that function was `localhost` - the value required by our tests. To solve this issue:

- 84b261a: A Lambda environment detection mechanism has been implemented by using [reserved environment variables](https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html)
- 4c3e476: Corresponding unit tests have been added

**Benchmark**
The numerous changes implemented by the current PR did not impact performance or error rate in high throughput scenarios:

v1.0.0
image

Current PR
image

axw (approved):
Thanks for the update! Change looks good.

PR #199, opened 2022-05-30
Added building and pushing docker image for the lambda-extension

### LINKED ISSUE #195: Publish extension as a docker image
Currently, we only support Elastic APM with Lambda function that are deployed as ZIP archives.
In order to support scenarios where users deploy their function as a docker container, we need to provide an easy way to integrate the lambda extension into their target Dockerfile.

For this we need to publish the lambda extension as a Docker image, using a simple Dockerfile like:

```
FROM scratch
ARG EXTENSION_BINARY
COPY ${EXTENSION_BINARY} /opt/extensions
```

The users can then use this image to easily include the extension in their Docker image for their function:

```
FROM docker.elastic.co/observability/lambda-extension-layer AS lambda-extension
FROM docker.elastic.co/observability/java-apm-layer AS java-agent

FROM public.ecr.aws/lambda/java:11

# Extension Layer code
COPY --from=lambda-extension /opt /opt               <-- COPY LAMBDA EXTENSION
COPY --from=java-agent /opt /opt                           <-- COPY JAVA AGENT

# Copy function code and runtime dependencies from Gradle layout
COPY build/classes/java/main ${LAMBDA_TASK_ROOT}
COPY build/dependency/* ${LAMBDA_TASK_ROOT}/lib/



# Set the CMD to your handler (could also be done as a parameter override outside of the Dockerfile)
CMD [ "co.elastic.test.lambda.ApiGatewayV1Handler" ]
```

AlexanderWert:
Closes #195

### DIFF
--- a/.ci/push_docker.sh
+++ b/.ci/push_docker.sh
@@ -1,0 +1,42 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+# This script is present on workers but may not be present in a development
+# environment.
+
+if [ ${WORKSPACE+x} ]  # We are on a CI worker
+then
+  source /usr/local/bin/bash_standard_lib.sh
+fi
+
+readonly RETRIES=3
+
+readonly DOCKER_REGISTRY_URL="$1"
+
+readonly DOCKER_IMAGE_NAME="$2"
+
+readonly DOCKER_TAG="$3"
+
+readonly DOCKER_PUSH_IMAGE="$DOCKER_REGISTRY_URL/$DOCKER_IMAGE_NAME:$DOCKER_TAG"
+
+readonly DOCKER_PUSH_IMAGE_LATEST="$DOCKER_REGISTRY_URL/$DOCKER_IMAGE_NAME:latest"
+
+echo "INFO: Pushing image $DOCKER_PUSH_IMAGE to $DOCKER_REGISTRY_URL"
+
+if [ ${WORKERS+x} ]  # We are on a CI worker
+then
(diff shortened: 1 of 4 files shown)


.ci/push_docker.sh
```diff
@@ -0,0 +1,42 @@
...
+
+readonly DOCKER_TAG="$3"
+
+readonly DOCKER_PUSH_IMAGE="$DOCKER_REGISTRY_URL/$DOCKER_IMAGE_NAME:$DOCKER_TAG"
+
+readonly DOCKER_PUSH_IMAGE_LATEST="$DOCKER_REGISTRY_URL/$DOCKER_IMAGE_NAME:latest"
+
+echo "INFO: Pushing image $DOCKER_PUSH_IMAGE to $DOCKER_REGISTRY_URL"
+
+if [ ${WORKERS+x} ]  # We are on a CI worker
```
v1v, line 27: I'm not much familiar with `WORKERS` but `CI` is the default env variable if it helps

v1v (commented):
LGTM, minor comments, only one question regarding the docker image for multiarch, as far as I see, the docker image will be different for x86 and arm. Is that the intended? I'm happy about this anyway, but wanted to double check if for any reasons we wanted a docker image multi-arch based instead, since it will require some other implementation

v1v (approved):

.ci/push_docker.sh
AlexanderWert: I copied this logic from the [Java agent's push_docker.sh](https://github.com/elastic/apm-agent-java/blob/main/scripts/jenkins/push_docker.sh)

PR #208, opened 2022-06-03
Added Secret Manager support

### LINKED ISSUE #105: Support for Secrets Managers
At present, APM Server credentials are stored in env. variables.  In addition to this standard method for supplying credentials we should investigate Secrets Managers ([such as AWS Secrets Manager](https://aws.amazon.com/secrets-manager/) and determine how feasible it would be to allows the APM Server credentials to be stored and fetched via a secrets manager API

Next steps

- [ ] General Investigation of AWS Secrets Manager (look for blockers, challenges, etc.)
- [ ] Other Secrets Managers?
- [ ] Write Extension code to fetch secrets and populated needed credentials
- [ ] Document use of the secrets manager for end users

### LINKED ISSUE #206: Normalize log level values
The extension currently only accepts correctly capitalized log level values.
For example, the value `DEBUG` is ignored (and it falls back to `trace`), because the extension expects the value `debug` instead.

We need to lowercase the log level values before using them.

AlexanderWert:
Closes #105
Closes #206

## Description
Added two new configuration options:
- `ELASTIC_APM_SECRETS_MANAGER_SECRET_TOKEN_ID`
- `ELASTIC_APM_SECRETS_MANAGER_API_KEY_ID`

If a secret manager id is provided through the above config options, the extension retrieves the corresponding secret from the AWS Secret manager. The secret must be in the same AWS region. 
If both, a plain text secret is provided (i.e. `ELASTIC_APM_SECRET_TOKEN`) and the `ELASTIC_APM_SECRETS_MANAGER_SECRET_TOKEN_ID`, then the value from the Secret Manager takes precedence.

Retrieving the secret from the Secret Manager happens once at the cold start of the extension.

### Step 1: Create a secret in the AWS Secret Manager.
- The secret must be in the same AWS region as the target lambda function
- Create the secret either for the APM secret token OR the APM API key
- The secret type **must be** `Plain Text`
- Optionally, you can select a custom KMS key for encryption (Note: in this case you will need additional key permissions on your Lambda functions)
  - Otherwise: use the key provided by AWS: `aws/secretsmanager` (No additional key permissions needed for this)
  
image

### Step 2: Create and setup the Lambda function
- Follow [these docs](https://www.elastic.co/guide/en/apm/guide/8.2/monitoring-aws-lambda.html) for general setup of the Lambda function with Elastic APM
- However, create and use a custom extension layer based on this PR
- In the environment variables configuration **instead** of providing the `ELASTIC_APM_SECRET_TOKEN` or `ELASTIC_APM_API_KEY` specify the `ELASTIC_APM_SECRETS_MANAGER_SECRET_TOKEN_ID` or `ELASTIC_APM_SECRETS_MANAGER_API_KEY_ID` variables. Use the `secret-name` as the value.

### Step 3: Add permissions
- Under `Configuration` > `Permissions` > `Execution Role` click on the execution role to edit the permissions
- Add a new inline policy with the following JSON-based permission definition:
```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "secretsmanager:GetSecretValue",
                "secretsmanager:DescribeSecret",
                "secretsmanager:ListSecretVersionIds"
            ],
            "Resource": [
                "HERE_GOES_THE_ARN_OF_THE_SECRET"
            ]
        },
        // The following is only needed if a custom key is used
        {
            "Effect": "Allow",
            "Action": [
                "kms:Decrypt"
            ],
            "Resource": [
                "HERE_GOES_THE_ARN_OF_THE_CUSTOM_KEY"
            ]
        }
    ]
}
```

### Step 4: Invoke the lambda function and validate data in Elastic APM UI


apm-lambda-extension/extension/process_env.go
```diff
@@ -23,6 +23,10 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"github.com/aws/aws-sdk-go/service/secretsmanager"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"encoding/base64"
```
stuartnelson3, line 29: is there any sort of linting for import order? other projects separate out stdlib / elastic / 3rd party

apm-lambda-extension/extension/process_env.go
```diff
@@ -62,6 +66,39 @@ func getIntFromEnv(name string) (int, error) {
 	return value, nil
 }
 
+func getSecret(secretName string) (string, error) {
+	region := os.Getenv("AWS_REGION")
```
stuartnelson3, line 70: might want to check region != "" and return a descriptive error

apm-lambda-extension/extension/process_env.go
```diff
@@ -62,6 +66,39 @@ func getIntFromEnv(name string) (int, error) {
...
+	if err != nil {
+		return "", err
+	}
+
+	var secretString string
+	if result.SecretString != nil {
+		secretString = *result.SecretString
+	} else {
+		decodedBinarySecretBytes := make([]byte, base64.StdEncoding.DecodedLen(len(result.SecretBinary)))
+		len, err := base64.StdEncoding.Decode(decodedBinarySecretBytes, result.SecretBinary)
```
stuartnelson3, line 92: `len` is a builtin and, while the compiler might let you use it, it'd probably be better to choose a different variable name

apm-lambda-extension/extension/process_env.go
```diff
@@ -95,10 +132,34 @@ func ProcessEnv() *extensionConfig {
...
+		}
+	}
+
+	apmServerSecretToken := os.Getenv("ELASTIC_APM_SECRET_TOKEN")
+	apmServerSecretTokenSMSecretId := os.Getenv("ELASTIC_APM_SECRETS_MANAGER_SECRET_TOKEN_ID")
+	if apmServerSecretTokenSMSecretId != "" {
+		result, err := getSecret(apmServerSecretTokenSMSecretId)
+		if err != nil {
+			Log.Fatalf("Failed loading APM Server Secret Token from Secrets Manager: %v", err)
+		} else {
```
stuartnelson3, line 153: same as above

simitt:
@elastic/apm-server could you please help pushing this over the finish line; I suggest for this PR rather than only reviewing, commit small code improvements directly to free up @AlexanderWert. 
Thanks Alex for this contribution.

apm-lambda-extension/extension/process_env_test.go
```diff
@@ -172,9 +179,52 @@ func TestProcessEnv(t *testing.T) {
...
 	}
-	config = ProcessEnv()
+	config = ProcessEnv(sm)
 	if config.LogLevel != zapcore.InfoLevel {
 		t.Log("Log level not set correctly")
 		t.Fail()
 	}
 }
+
+func TestGetSecretCalled(t *testing.T) {
```
jlvoiseux, line 189: It seems that the Unit test, when run sequentially together with the whole test suite, fails due to previously set values of the `SECRET_TOKEN` and the `API_KEY`.
```suggestion
func TestGetSecretCalled(t *testing.T) {
    os.Clearenv()
```

jlvoiseux (changes requested):
LGTM, aside from the small test correction.
I was also able to get the PR to work end-to-end by following the repro steps.

PR #207, opened 2022-06-03
Generate NOTICE.txt and dependencies.asciidoc

stuartnelson3:
- add script to generate NOTICE.txt and dependencies.asciidoc
- generate NOTICE.txt and dependencies.asciidoc
- check they are up-to-date in ci

closes #197

PR #212, opened 2022-06-03
Add NOTICE.txt file to zip

simitt:
Add the `NOTICE.txt` file to the `zip` from which the ARN is built.


AlexanderWert (approved):

simitt:
depends on https://github.com/elastic/apm-aws-lambda/pull/207

PR #214, opened 2022-06-03
Update license headers for java, js, python

simitt:
Updating the Makefile to also add the license headers to the test examples written in `python`, `java` and `js`. 

* python: cannot be automated yet, due to https://github.com/elastic/go-licenser/issues/43
* java: automated
* js: adding the license header cannot be automated as the formatting would be changed, but the check can be automated and is done

### DIFF
--- a/apm-lambda-extension/Makefile
+++ b/apm-lambda-extension/Makefile
@@ -24,10 +24,12 @@ all: build
 check-licenses:
 	go install github.com/elastic/go-licenser@v0.4.0
 	go run github.com/elastic/go-licenser@v0.4.0 -d .
+	go run github.com/elastic/go-licenser@v0.4.0 -d -ext .java .
 
 update-licenses:
 	go install github.com/elastic/go-licenser@v0.4.0
 	go run github.com/elastic/go-licenser@v0.4.0 .
+	go run github.com/elastic/go-licenser@v0.4.0 -ext .java .
 
 lint:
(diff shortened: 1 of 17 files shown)


apm-lambda-extension/Makefile
```diff
@@ -24,10 +24,14 @@ all: build
...
 	go install github.com/elastic/go-licenser@v0.4.0
 	go run github.com/elastic/go-licenser@v0.4.0 -d .
+	go run github.com/elastic/go-licenser@v0.4.0 -d -ext .java .
+	go run github.com/elastic/go-licenser@v0.4.0 -d -ext .js .
 
 update-licenses:
 	go install github.com/elastic/go-licenser@v0.4.0
 	go run github.com/elastic/go-licenser@v0.4.0 .
+	go run github.com/elastic/go-licenser@v0.4.0 -ext .java .
+	go run github.com/elastic/go-licenser@v0.4.0 -ext .js .
```
AlexanderWert, line 34: you wrote in the description:
> js: adding the license header cannot be automated as the formatting would be changed, but the check can be automated and is done

Isn't this line adding / updating the license headers for .js files?

AlexanderWert (approved):

apm-lambda-extension/Makefile
simitt: Removed that line now. It changes the formatting/intentation on the .js files. I actually updated the headers for js files manually and then run the control script via the `make check-licenses`.

PR #216, opened 2022-06-09
Only add executable to extensions/

axw:
The extensions directory must only contain executable extensions.
Add NOTICE.txt and dependencies.asciidoc to the top level of the
zip, rather than in the extensions directory.


lahsivjar (approved):
LGTM

Issue #205, opened 2022-06-03
Log a warning when authentication with APM server fails


AlexanderWert:
Currently, when the extension fails to authenticate with the APM server, extension does not log any warning.
Only when switching log level to `debug`, we see that the response body from the APM server contains the following: 
<img width="725" alt="image" src="https://user-images.githubusercontent.com/866830/171799339-26dab0e3-c47b-42c6-87ee-e3f8b2bd8477.png">

Nevertheless, the APM transport is set to `healthy` state (which is not correct!).

We need explicit handling and logging in the extension when APM server authentication fails.

Issue #225, opened 2022-06-21
Parse APM Server response and react to error codes


simitt:
Parse APM Server response status code. Issue a log entry at the `error` log level if the `status_code >= 400`. Also adapt the `Health` status accordingly. 

https://github.com/elastic/apm-aws-lambda/blob/main/apm-lambda-extension/extension/apm_server_transport.go#L188-L191

PR #228, opened 2022-07-13
feat: log a warning when authentication with APM server fails 

kruskall:
The intake route of the APM server sends a 202 status accepted
if the request succeeds.

See https://github.com/elastic/apm-server/blob/main/docs/api-events.asciidoc#response

If auth fails the server returns a 401.

See https://github.com/elastic/apm-server/blob/3cd1a63bf218ea0ab0f240924236823e1815882b/beater/middleware/auth_middleware.go#L54-L57 and https://github.com/elastic/apm-server/blob/3cd1a63bf218ea0ab0f240924236823e1815882b/beater/request/result.go#L86

Related to https://github.com/elastic/apm-aws-lambda/issues/205+

### DIFF
--- a/apm-lambda-extension/extension/apm_server_transport.go
+++ b/apm-lambda-extension/extension/apm_server_transport.go
@@ -185,10 +185,10 @@ func (transport *ApmServerTransport) PostToApmServer(ctx context.Context, agentD
 		return fmt.Errorf("failed to read the response body after posting to the APM server")
 	}
 
-	// On success, the server will respond with a 202 Accepted status code.
-	// Log a warning otherwise.
-	if resp.StatusCode != http.StatusAccepted {
-		Log.Warnf("APM server request failed with status code: %d", resp.StatusCode)
+	if resp.StatusCode == http.StatusUnauthorized {
+		Log.Warnf("Authentication with the APM server failed: response status code: %d", resp.StatusCode)
+		Log.Debugf("APM server response body: %v", string(body))
+		return nil
 	}
 
 	transport.SetApmServerTransportState(ctx, Healthy)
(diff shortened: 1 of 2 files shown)


apm-lambda-extension/extension/apm_server_transport.go
```diff
@@ -185,6 +185,12 @@ func (transport *ApmServerTransport) PostToApmServer(ctx context.Context, agentD
 		return fmt.Errorf("failed to read the response body after posting to the APM server")
 	}
 
+	if resp.StatusCode == http.StatusUnauthorized {
+		Log.Warnf("Authentication with the APM server failed: response status code: %d", resp.StatusCode)
+		Log.Debugf("APM server response body: %v", string(body))
+		return nil
```
kruskall, line 191: **question:** should this return an error or is it too disruptive ?

simitt: As you pointed out, #225  is a generalization where also the health state should be adapted for various scenarios. This will require a slight rework of the [healthstate handling](https://github.com/elastic/apm-aws-lambda/pull/228/files#diff-84dece7b8082ce96bbb9405a021562cc58b344c8558ff8d38816f564c372c9e5R208-R235). I don't mind if you want to merge this improvement separately as is now and then create a follow up PR or add the health state handling and checking if there are other cases that need to be handled in this PR.

simitt: That's fine, just please remove this from the description and only close the issue when the health state is also fixed. 

> Closes https://github.com/elastic/apm-aws-lambda/issues/205

simitt (approved):

apm-lambda-extension/extension/apm_server_transport.go
kruskall: Done!

PR #233, opened 2022-07-14
refactor: use native test env variable

kruskall:
Go 1.17 added support for setting an environment variable for the
duration of test.
t.Setenv provides automatic cleanup of env variables after the
test ends.

See https://go.dev/doc/go1.17#testing


marclop (approved):
Looks good!

PR #234, opened 2022-07-14
fix: properly handle http.ErrServerClosed

kruskall:
Logs API server: the server was ignoring ErrServerClosed and logging an error.

APM Data server: the server was using an equality check with the error message but
it was never satisfied because of a typo ('server closed' vs 'Server closed'), leading
to an error message.

Both servers are now handling the error and logging a message that the server stopped
successfully. The code is using go 1.13 errors to avoid comparing strings and for better
compatibility.

### DIFF
--- a/apm-lambda-extension/extension/http_server.go
+++ b/apm-lambda-extension/extension/http_server.go
@@ -19,6 +19,7 @@ package extension
 
 import (
 	"context"
+	"errors"
 	"net"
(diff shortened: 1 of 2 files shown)


marclop (approved):
Looks good! Just left a couple of suggestions on rewording the errors and using a non-capitalized error to comply with static analysis tools.

PR #237, opened 2022-07-15
feat: add support for reproducible builds

kruskall:
Add support for SOURCE_DATE_EPOCH to avoid timestamp and timezones
issues.

See https://reproducible-builds.org/specs/source-date-epoch/

Make sure zip entries have a consistent mtime and strip extra
attributes.

See https://reproducible-builds.org/docs/archives/

Zip archives are now reproducible by default. Update script
to use the make task and avoid calling zip directly.

For more information see https://reproducible-builds.org/

### DIFF
--- a/apm-lambda-extension/Makefile
+++ b/apm-lambda-extension/Makefile
@@ -7,7 +7,7 @@ AGENT_VERSION = $(shell echo $${BRANCH_NAME} | cut -f 2 -d 'v')
 
 # Add support for SOURCE_DATE_EPOCH and reproducble buils
 # See https://reproducible-builds.org/specs/source-date-epoch/
-SOURCE_DATE_EPOCH ?= 0
+SOURCE_DATE_EPOCH ?= $(shell git log -1 --pretty=%ct)
 
 ifndef GOARCH
 	GOARCH=amd64
--- a/apm-lambda-extension/Makefile
+++ b/apm-lambda-extension/Makefile
@@ -62,7 +62,7 @@ endif
 	$(MAKE) publish
 zip:
 	cd bin \
-	&& rm -f extension.zip || true \
+	&& rm -f extension.zip \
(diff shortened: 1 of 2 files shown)


apm-lambda-extension/Makefile
```diff
@@ -57,7 +61,11 @@ endif
 	GOARCH=${GOARCH} make zip
 	$(MAKE) publish
 zip:
-	cd bin && rm -f extension.zip || true && zip -r extension.zip extensions NOTICE.txt dependencies.asciidoc && cp extension.zip ${GOARCH}.zip
+	cd bin \
+	&& rm -f extension.zip || true \
```
axw, line 65:
```suggestion
	&& rm -f extension.zip \
```

pretty sure that's redundant with the `-f` flag?

apm-lambda-extension/Makefile
```diff
@@ -5,6 +5,10 @@ DOCKER_IMAGE_NAME = observability/apm-lambda-extension
 DOCKER_REGISTRY = docker.elastic.co
 AGENT_VERSION = $(shell echo $${BRANCH_NAME} | cut -f 2 -d 'v')
 
+# Add support for SOURCE_DATE_EPOCH and reproducble buils
+# See https://reproducible-builds.org/specs/source-date-epoch/
+SOURCE_DATE_EPOCH ?= 0
```
axw, line 10: Should we default to using the latest commit timestamp (`$(git log -1 --pretty=%ct)`)?

axw (commented):
Nice, I wasn't aware of that spec. Looks good, just wondering if we should set a different default in the Makefile.

apm-lambda-extension/Makefile
kruskall: I think this makes sense and the spec seems to allow it:

> The value MUST be reproducible (deterministic) across different executions of the build, depending only on the source code. It SHOULD be set to the last modification time of the source, incorporating any packaging-specific modifications.

Issue #188, opened 2022-04-21
Revisit backoff grace period computation


jlvoiseux:
The current exponential backoff mechanism can lead to multiple Lambda environments being synchronously stuck, as they will all attempt to try to reconnect to a potentially unresponsive APM Server at the same time. 

The following picture is a bar chart illustrating the duration of several invocations over 10 minutes, coupled with a slow APM Server. There is a very visible synchronous pattern of multiple Lambda environments stuck trying to query the APM Server at the same time, leading to the creation of new environments and increasing the cold start rate.

![image](https://user-images.githubusercontent.com/48380853/164467516-92dadd0a-b10d-45ac-8707-3210052701db.png)

Immediate mitigation actions to be attempted are:

- [ ] Increase the initial grace period, to avoid the creation of a lot of Lambda environments at the start
- [ ] Increase the jitter to make the grace period duration more random, and assign a specific jitter value to each instance of the extension at cold start
- [ ] Perform a high-throughput benchmark over a long period of time to monitor the environment scaling.

PR #240, opened 2022-07-17
fix: randomise the initial grace period to avoid collisions

kruskall:
The previous algorithm was using binary exponential-backoff with
a +- 10% jitter to calculate the grace period.
Because there can be multiple lambda environments we need to
mitigate collisions:

We cannot use 0 as the first delay because functions failing closer to
each other will collide. The issue would then be propagated by the
small jitter for lower delays.

This change adds an initial delay of n seconds to the first reconnection
attempt.
n is randomly generated in a closed interval to account for collisions
while keeping in mind usability and user experience.

Closes https://github.com/elastic/apm-aws-lambda/issues/188

Potential followup issue: make the interval configurable with an environment variable


apm-lambda-extension/extension/apm_server_transport.go
```diff
@@ -232,6 +232,13 @@ func (transport *ApmServerTransport) SetApmServerTransportState(ctx context.Cont
 
 // ComputeGracePeriod https://github.com/elastic/apm/blob/main/specs/agents/transport.md#transport-errors
 func (transport *ApmServerTransport) computeGracePeriod() time.Duration {
+	// If reconnectionCount is 0, returns a random number in an interval.
+	// The grace period for the first reconnection count was 0 but that
+	// leads to collisions with multiple environment.
+	if transport.reconnectionCount == 0 {
+		gracePeriod := rand.Float64() * 5
```
marclop, line 239: Should seed the rand library via `rand.Seed()` or create a new `*rand.Rand` and use it in ApmServerTransport. WDYT?

kruskall: Sorry, I missed this comment :bow: 

That's a great idea! I've seeded the rand library :+1:

simitt (approved):

PR #244, opened 2022-07-18
refactor: move main loop to a separate package and avoid global state

kruskall:
Create an app struct to hold the state of the application and
use functional options for flexibility.
Remove global extension client and avoid mixing test code
with app logic.

This is part of a bigger effort to improve testing and
decouple different components.


axw (approved):
Looks great!

PR #248, opened 2022-07-22
refactor: redesign logs api

kruskall:
Use a logsapi client and add functional options to make it more
flexible in local or testing environments.

Add support for graceful shutdown to the HTTP server and only stop it
after the event has been processed to avoid race conditions on the log
channel.

Add more detailed errors and make sure they are wrapped to provide
appropriate context.

Rename and move code around for clarity.


apm-lambda-extension/logsapi/subscribe_test.go
```diff
@@ -1,136 +0,0 @@
...
-}
-
-func TestSubscribeWithLambdaFunction(t *testing.T) {
-	t.Setenv("AWS_LAMBDA_FUNCTION_NAME", "mock")
-
-	_, err := Subscribe(context.Background(), "testID", []EventType{Platform})
-	assert.Error(t, err, "listen tcp: lookup sandbox: no such host")
-}
-
-func TestSubscribeAWSRequest(t *testing.T) {
```
axw, line 46: I don't see any equivalent of this test. Is that on purpose? Seems like a useful test, ensuring the low level log event delivery works.

I wonder if we should move the `ProcessLogs` and `ProcessPlatformReport` methods out of this package, and into either the `app` or `extension` package? Those methods are business logic, and I feel like it would be a better separation if we made `logsapi` just about delivering the raw log events. That would also make it more easily testable.

apm-lambda-extension/app/run.go
```diff
@@ -66,9 +66,16 @@ func (app *App) Run(ctx context.Context) error {
 	// Use a wait group to ensure the background go routine sending to the APM server
 	// completes before signaling that the extension is ready for the next invocation.
 
-	logsTransport, err := logsapi.Subscribe(ctx, app.extensionClient.ExtensionID, []logsapi.EventType{logsapi.Platform})
-	if err != nil {
+	if err := app.logsClient.StartService([]logsapi.EventType{logsapi.Platform}, app.extensionClient.ExtensionID); err != nil {
```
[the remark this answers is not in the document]
kruskall: Yes, good catch! :+1:

apm-lambda-extension/app/run.go
```diff
@@ -81,10 +88,18 @@ func (app *App) Run(ctx context.Context) error {
 		select {
 		case <-ctx.Done():
 			extension.Log.Info("Received a signal, exiting...")
+
+			// Remember to shutdown the log service if available.
+			if app.logsClient != nil {
+				if err := app.logsClient.Shutdown(); err != nil {
+					extension.Log.Warnf("failed to shutdown the log service: %v", err)
+				}
+			}
```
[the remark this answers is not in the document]
kruskall: Yep, I missed the other return, I think a defer is safer! :+1:

apm-lambda-extension/app/run.go
```diff
@@ -66,9 +66,16 @@ func (app *App) Run(ctx context.Context) error {
...
 
-	logsTransport, err := logsapi.Subscribe(ctx, app.extensionClient.ExtensionID, []logsapi.EventType{logsapi.Platform})
-	if err != nil {
+	if err := app.logsClient.StartService([]logsapi.EventType{logsapi.Platform}, app.extensionClient.ExtensionID); err != nil {
 		extension.Log.Warnf("Error while subscribing to the Logs API: %v", err)
+
+		// remember to shutdown the service if started
+		if err := app.logsClient.Shutdown(); err != nil {
+			extension.Log.Warnf("failed to shutdown the log service: %v", err)
+		}
```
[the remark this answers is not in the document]
kruskall: Agreed, we also cannot assume the server started if we got an error! Changed :+1:

apm-lambda-extension/logsapi/client_test.go
```diff
@@ -0,0 +1,99 @@
...
+		t.Run(name, func(t *testing.T) {
+			_, err := logsapi.NewClient(tc.opts...)
+			if tc.expectedErr {
+				require.Error(t, err)
+			}
+		})
+	}
+}
+
+func TestSubscribe(t *testing.T) {
```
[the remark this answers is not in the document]
kruskall: Great point! Added!

apm-lambda-extension/logsapi/subscribe_test.go
kruskall: Yep, the logs listener is not tested. I've added a new test to make sure it is handling requests properly.

> I wonder if we should move the ProcessLogs and ProcessPlatformReport methods out of this package, and into either the app or extension package? Those methods are business logic, and I feel like it would be a better separation if we made logsapi just about delivering the raw log events. That would also make it more easily testable.

I agree with you there, that's probably the last thing preventing a proper decoupling between logsapi and extension package. I think we should move those methods but I wanted to keep the focus of this PR to the `logsapi` package, without touching external code.

Let me know if you want to move them in this PR!

apm-lambda-extension/main_test.go
```diff
@@ -345,13 +344,14 @@ func sendLogEvent(requestId string, logEventType logsapi.SubEventType) {
 		extension.Log.Errorf("Could not encode record : %v", err)
 		return
 	}
-	host, port, _ := net.SplitHostPort(logsapi.TestListenerAddr.String())
+	// TODO refactor these tests
```
axw, line 347: Did you say you were going to skip the tests until this was done? The tests in the main package are failing at the moment because of this.

apm-lambda-extension/logsapi/subscribe_test.go
axw: OK, sounds good - I'm happy for that to be done in a followup.

axw (commented):
Thanks @kruskall, those changes look good. I'm happy for the rest to be done in followup PRs.

apm-lambda-extension/main_test.go
kruskall: Done: https://github.com/elastic/apm-aws-lambda/issues/249

kruskall:
Thank you for the feedback! :bow:

axw (approved):
LGTM, thanks :)

PR #250, opened 2022-07-26
refactor: abstract apm client to a separate package

kruskall:
Start working on decoupling apm and extension package.
Shutdown the apm receiver gracefully and move away from
deprecated ioutil calls.

Still untested but ready for review


apm-lambda-extension/apm/apmserver.go
```diff
@@ -0,0 +1,214 @@
...
+	"errors"
+	"fmt"
+	"io"
+	"math"
+	"math/rand"
+	"net/http"
+	"time"
+)
+
+// StartBackgroundApmDataForwarding Receive agent data as it comes in and post it to the APM server.
```
axw, line 34:
```suggestion
// ForwardApmData receives agent data as it comes in and posts it to the APM server.
```

axw (approved):
Thanks, loving this refactoring.

Just one minor thing: "apm" feels too generic of a package name. Perhaps call it "apmproxy", or something along those lines?

apm-lambda-extension/apm/apmserver.go
kruskall: Good catch! 

I also renamed the package :+1:

PR #251, opened 2022-07-28
feat: move logger to a separate package

kruskall:
This is the last step to eliminate global state. Moreover,
'apmproxy' and 'extension' are now completely separate and
do not depend on each other.


apm-lambda-extension/extension/process_env.go
```diff
@@ -101,7 +103,7 @@ func ptrFromString(v string) *string {
 }
 
 // ProcessEnv extracts ENV variables into globals
-func ProcessEnv(manager secretManager) *extensionConfig {
+func ProcessEnv(manager secretManager, Log *zap.SugaredLogger) *extensionConfig {
```
axw, line 106:
```suggestion
func ProcessEnv(manager secretManager, log *zap.SugaredLogger) *extensionConfig {
```

(and rename references to Log below)

apm-lambda-extension/apmproxy/receiver_test.go
```diff
@@ -48,13 +49,17 @@ func TestInfoProxy(t *testing.T) {
 	}))
 	defer apmServer.Close()
 
+	l, err := logger.New()
+	require.NoError(t, err)
```
axw, line 53: Instead of repeating this in all the tests, can we use [`zaptest.NewLogger`](https://pkg.go.dev/go.uber.org/zap@v1.21.0/zaptest#NewLogger)? That could be used inline in the `WithLogger` call to keep the tests a little bit neater.

axw (approved):
:tada: 

LGTM, but maybe we can cut down some of the repetition by using zaptest for constructing a logger. In the tests we can just panic/fail the test if the logger construction fails for some unexpected reason.

apm-lambda-extension/extension/process_env.go
kruskall: Good catch!

apm-lambda-extension/apmproxy/receiver_test.go
kruskall: TIL about `zaptest`! It looks much cleaner now 👍

PR #252, opened 2022-07-28
fix: handle main loop errors correctly

kruskall:
Do not call fatal or panic directly, instead propagate the error
to avoid multiple point of failure.


axw (approved):
LGTM!

PR #253, opened 2022-07-28
fix: do not enqueue broken metrics data to the apm server

kruskall:
If an error is returned when processing platform metrics, the
processed metrics are enqueued to the apm server regardless.

We can't assume anything about the processed metrics so we
return an error which is logged correctly.

### DIFF
--- a/apm-lambda-extension/logsapi/event.go
+++ b/apm-lambda-extension/logsapi/event.go
@@ -93,10 +93,10 @@ func (lc *Client) ProcessLogs(
 					extension.Log.Debug("Received platform report for the previous function invocation")
 					processedMetrics, err := ProcessPlatformReport(metadataContainer, prevEvent, logEvent)
 					if err != nil {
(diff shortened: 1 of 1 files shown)


axw:
> If an error is returned when processing platform metrics, the processed metrics are enqueued to the apm server regardless.

I don't think that's the case. We were logging the error, and only calling EnqueueAPMData in the non-error case. We would previously continue processing log events in the case where ProcessPlatformReport returned an error.

kruskall:
> > If an error is returned when processing platform metrics, the processed metrics are enqueued to the apm server regardless.
> 
> I don't think that's the case. We were logging the error, and only calling EnqueueAPMData in the non-error case. We would previously continue processing log events in the case where ProcessPlatformReport returned an error.

uh!? I'm really sorry about this, the correct PR is https://github.com/elastic/apm-aws-lambda/pull/257. I'll close this one and add a reminder to myself to get enough sleep :sweat_smile:

axw:
No worries :)

PR #255, opened 2022-07-28
fix: retract version 1.0.1

kruskall:
v1.0.1 is broken and should not be used. Add an explicit
retract directive to the go.mod.

See https://go.dev/doc/modules/gomod-ref#retract


axw (approved):
LGTM, but is this necessary/useful for something that's not a library?

kruskall:
Looking into this a bit more I don't think it's really necessary. v1.0.1 is not technically broken, the release zip is.

Using the docker image or building the extension into a docker image would work just fine.

I'll close this. Thanks for the help :bow:

Issue #245, opened 2022-07-18
Server panic when sending events to APM server


richcto:
Extension version 1.1.0, elastic-apm version 6.10.0

Even though all errors are sent to Elastic APM endpoint, I still get errors logged in lambda log group. See below:

`

{     "log.level": "info",     "@timestamp": "2022-07-18T17:17:41.973Z",     "log.origin": {         "file.name": "extension/http_server.go",         "file.line": 48     },     "message": "Extension listening for apm data on :8200",     "ecs.version": "1.6.0" }
--
{     "log.level": "info",     "@timestamp": "2022-07-18T17:17:43.124Z",     "log.origin": {         "file.name": "logsapi/subscribe.go",         "file.line": 123     },     "message": "Extension listening for Lambda Logs API events on 169.254.79.130:42187",     "ecs.version": "1.6.0" }
LOGS	Name: elastic-apm-extension	State: Subscribed	Types: [platform]
{     "log.level": "info",     "@timestamp": "2022-07-18T17:17:43.277Z",     "log.origin": {         "file.name": "apm-lambda-extension/main.go",         "file.line": 129     },     "message": "Waiting for next event...",     "ecs.version": "1.6.0" }
START RequestId: 3dccf0cb-7d34-4cfd-b73e-f562ad914351 Version: $LATEST
EXTENSION	Name: elastic-apm-extension	State: Ready	Events: [SHUTDOWN,INVOKE]
{"time_stamp": "2022-07-18 17:17:51,677","log_level": "INFO","log_message": Processing event: {'key1': 'value1', 'key2': 'value2', 'key3': 'value3'}}
{"time_stamp": "2022-07-18 17:17:51,678","log_level": "INFO","log_message": Initialising for event: {'key1': 'value1', 'key2': 'value2', 'key3': 'value3'}}
2022/07/18 17:17:51 http: panic serving 127.0.0.1:58162: send on closed channel
goroutine 40 [running]:
net/http.(*conn).serve.func1()
/usr/local/Cellar/go/1.17.6/libexec/src/net/http/server.go:1802 +0xb9
panic({0x7948e0, 0x87ea20})
/usr/local/Cellar/go/1.17.6/libexec/src/runtime/panic.go:1047 +0x266
elastic/apm-lambda-extension/extension.handleIntakeV2Events.func1({0x88acf0, 0xc00032c700}, 0xc0000f2f00)
[...]

PR #258, opened 2022-08-01
fix: always flush data to apm before shutting down and rework agent done signal

kruskall:
Add a defer statement to make sure that we always flush data to the
apm server before shutting down.

Remove agent done signal channel and avoid leaking implementation details.
The channel was being recreated and closed on each event, racing with the
intake handler that was sending to the channel.
The channel is now used internally by the apm client and external packages
can call 'Done()' to check whether the agent has sent the final intake
request.

See https://github.com/elastic/apm/blob/main/specs/agents/tracing-instrumentation-aws-lambda.md#data-flushing

Closes https://github.com/elastic/apm-aws-lambda/issues/245


apm-lambda-extension/apmproxy/receiver.go
```diff
@@ -131,7 +131,7 @@ func (c *Client) handleIntakeV2Events() func(w http.ResponseWriter, r *http.Requ
 		}
 
 		if len(r.URL.Query()["flushed"]) > 0 && r.URL.Query()["flushed"][0] == "true" {
-			c.AgentDoneSignal <- struct{}{}
+			close(c.done)
```
axw, line 134: This doesn't look right: wouldn't this potentially lead to closing the channel multiple times (and panicking) when there are multiple Lambda invocations for the same extension lifetime? The extension may outlive a single invocation, i.e. on a warm start. In each invocation the APM agent may send a `?flushed=true` request to the extension.

kruskall: Thanks for spotting this! I've updated the code

apm-lambda-extension/app/run.go
```diff
@@ -201,9 +207,12 @@ func (app *App) processEvent(
 	// 2) [Backup 1] RuntimeDone is triggered upon reception of a Lambda log entry certifying the end of the execution of the current function
 	// 3) [Backup 2] If all else fails, the extension relies of the timeout of the Lambda function to interrupt itself 100 ms before the specified deadline.
 	// This time interval is large enough to attempt a last flush attempt (if SendStrategy == syncFlush) before the environment gets shut down.
+
+	if app.apmClient.HasPendingFlush() {
```
axw, line 211: Doesn't this mean that if the the agent sends a `...?flushed=true` request to the extension a bit too late (i.e. after this call to HasPendingFlush), then the extension will wait until the runtimeDone or timeout events have occurred? I don't think that's desirable - the extension is meant to continue as soon as it can, to minimise latency overhead.

kruskall: Good catch! I've tried a different solution, let me know what you think!

apm-lambda-extension/apmproxy/apmserver.go
```diff
@@ -210,11 +222,46 @@ func (c *Client) ComputeGracePeriod() time.Duration {
...
+		return true
 	default:
 		c.logger.Warn("Channel full: dropping a subset of agent data")
+		return false
 	}
 }
+
+// ShouldFlush returns true if the client should flush APM data after processing the event.
+func (c *Client) ShouldFlush() bool {
+	return c.sendStrategy == SyncFlush || c.hasPendingFlush()
```
axw, line 238: Looking at this again, I don't know that this is desirable.

I think the `?flushed=true` wording is a bit confusing: it doesn't mean that the extension should flush immediately, it just means that the agent (client) has flushed, which in turn means that the Lambda invocation has completed.

I think we should revert to only synchronously flushing when sendStrategy == SyncFlush.

apm-lambda-extension/apmproxy/apmserver.go
```diff
@@ -210,11 +222,46 @@ func (c *Client) ComputeGracePeriod() time.Duration {
...
+}
+
+// WaitForFlush returns a channel that is closed if the client has received a signal to flush
+// the buffered APM data.
+func (c *Client) WaitForFlush(ctx context.Context) <-chan struct{} {
+	ch := make(chan struct{})
+	go func() {
+		t := time.NewTimer(10 * time.Millisecond)
```
axw, line 246: This approach solves the latency issue, so it's definitely better - but I am concerned about busy-waiting causing avoidable CPU overhead. e.g. imagine if the Lambda invocation runs for seconds. I won't block the PR on this as this is better than panicking, but we should follow up on this ASAP.

I wonder if we might be better off with a control loop like in the Go agent, which always runs in the background and receives commands to flush, like: https://github.com/elastic/apm-agent-go/blob/83d9a9518d5e956d04e750a3bae5cbd0d58f682a/tracer.go#L589-L603. That's probably too much of a change for now though.

This would also probably all be simpler if we could pair the `?flushed` requests up with specific Lambda invocation request IDs.

apm-lambda-extension/apmproxy/apmserver.go
```diff
@@ -43,6 +43,12 @@ func (c *Client) ForwardApmData(ctx context.Context, metadataContainer *Metadata
 			c.logger.Debug("Invocation context cancelled, not processing any more agent data")
 			return nil
 		case agentData := <-c.DataChannel:
+			if agentData.Flushed {
+				c.flushMutex.Lock()
+				c.flushCount--
+				c.flushMutex.Unlock()
+				continue
```
axw, line 50: Why are we continuing here? IIANM, the `?flushed` request may contain event data.

apm-lambda-extension/apmproxy/apmserver.go
```diff
@@ -235,11 +235,32 @@ func (c *Client) EnqueueAPMData(agentData AgentData) bool {
...
+// WaitForFlush returns a channel that is closed if the client has received a signal to flush
+// the buffered APM data.
+func (c *Client) WaitForFlush(ctx context.Context) <-chan struct{} {
+	ch := make(chan struct{})
+	go func() {
+		t := time.NewTimer(10 * time.Millisecond)
+		defer t.Stop()
+		for {
+			select {
+			case <-ctx.Done():
```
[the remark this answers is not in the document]
kruskall: This should be handled by the new method not relying on ctx

apm-lambda-extension/apmproxy/apmserver.go
kruskall: I think that's a valid concern, I've removed the busy loop and the code is entirely relying on channels now. It is similar to the original approach before this PR but should handle multiple `flushed=true` requests and avoid race conditions

apm-lambda-extension/apmproxy/apmserver.go
kruskall: Good catch! Fixed

apm-lambda-extension/apmproxy/apmserver.go
kruskall: I might be misinterpreting the specification, but I think the goal of `flushing=true` was to reduce latency so that the lambda knows that it can flush its data.

The way I interpreted it was:
- `SyncFlush`: flush on every intake request. We forwards every request from the agent as soon as we receive it.
- `flushed=true`: sent with the final intake request. We buffered the previous requests and this is a signal that we can flush the data.

> I think we should revert to only synchronously flushing when sendStrategy == SyncFlush.

Question: Wouldn't that mean that buffered data have a chance of being flushed only on shutdown or while processing an event ? Is that intended ?
I think that would lead to a potential delay since if the buffer is not emptied we would have to wait for shutdown which could take a while.

axw: This is why I think the naming is confusing :)

There's two distinct "flush" events:
 1. agent flushes data to the extension
 2. extension flushes data to the server

The `?flushed=true` request indicates to the extension that the first event has happened. This is a _prerequisite_ for the extension flushing data to the server, but it does not mean the server must flush immediately. This behaviour is intended to be controlled by the [send strategy](https://github.com/elastic/apm-aws-lambda/blob/main/docs/monitoring-aws-lambda.asciidoc#elastic_apm_send_strategy).

> Question: Wouldn't that mean that buffered data have a chance of being flushed only on shutdown or while processing an event ? Is that intended ?
> I think that would lead to a potential delay since if the buffer is not emptied we would have to wait for shutdown which could take a while.

Yes, that is intended. There's a trade-off, as explained in the send strategy docs linked above:
 - In `syncflush` mode, data gets flushed immediately after an invocation. This means the extension cannot service another Lambda invocation until the events are flushed, reducing Lambda invocation throughput.
 - In `background` mode, data gets flushed in the background while subsequent Lambda invocations are being processed, or on shutdown if there are no subsequent invocations. This means that data might be significantly delayed in case there are only sporadic Lambda invocations, but Lambda invocation throughput will not be significantly reduced.

apm-lambda-extension/apmproxy/apmserver.go
axw: Thanks! I _think_ it's safe now. I also think we should look at reworking the flushing/concurrency in the extension in the not too distant future.

apm-lambda-extension/apmproxy/apmserver.go
kruskall: Ah I see, thank you for the explanation! 🙇 

I've updated the code to the correct behaviour

apm-lambda-extension/apmproxy/receiver.go
```diff
@@ -121,17 +122,36 @@ func (c *Client) handleIntakeV2Events() func(w http.ResponseWriter, r *http.Requ
...
+			Data:            rawBytes,
+			ContentEncoding: r.Header.Get("Content-Encoding"),
+			Flushed:         flushed,
 		}
 
-		if len(r.URL.Query()["flushed"]) > 0 && r.URL.Query()["flushed"][0] == "true" {
-			c.AgentDoneSignal <- struct{}{}
+		enqueued := c.EnqueueAPMData(agentData)
+
+		if enqueued && flushed {
```
axw, line 135: I missed this before: this means that we will not end `processEvent` early if the agent has indicated it has flushed, but the extension's queue is full. That doesn't seem ideal.

Perhaps instead of introducing `AgentData.Flushed` and recreating the channel when processing data, we could add a method `Client.ResetFlushed()`:

```go
// ResetFlushed resets the client's "agent flushed" state, such that
// subsequent calls to WaitForFlush will block until another request
// is received from the agent indicating it has flushed.
func (c *Client) ResetFlushed() {
    c.flushMutex.Lock()
    defer c.flushMutex.Unlock()
    c.flushCh = make(chan struct{})
}
```

Then we would call that at the end of `App.processEvent`? WDYT?

It's still not perfect, becaause `App.processEvent` could end due to the not-quite-timeout being reached. It's still possible that that timer fires but a `?flushed=true` request is then received, closing the channel after it's reset. I think we'll need to make more radical changes to fix that - later.

kruskall: Agree, we can focus on the linked bug here and tackle the other issues in a followup discussion//PR.

I liked `ResetFlushed`, I've updated the code

apm-lambda-extension/apmproxy/receiver.go
```diff
@@ -121,17 +121,32 @@ func (c *Client) handleIntakeV2Events() func(w http.ResponseWriter, r *http.Requ
...
+		agentData := AgentData{
+			Data:            rawBytes,
+			ContentEncoding: r.Header.Get("Content-Encoding"),
 		}
 
-		if len(r.URL.Query()["flushed"]) > 0 && r.URL.Query()["flushed"][0] == "true" {
-			c.AgentDoneSignal <- struct{}{}
+		enqueued := c.EnqueueAPMData(agentData)
+
+		if enqueued && flushed {
```
axw, line 133:
```suggestion
		if agentFlushed {
```

Closing the channel shouldn't depend on any enqueuing.

apm-lambda-extension/app/run.go
```diff
@@ -201,9 +207,13 @@ func (app *App) processEvent(
...
 	// This time interval is large enough to attempt a last flush attempt (if SendStrategy == syncFlush) before the environment gets shut down.
+
 	select {
-	case <-app.apmClient.AgentDoneSignal:
-		app.logger.Debug("Received agent done signal")
+	case <-app.apmClient.WaitForFlush():
+		app.logger.Debug("APM client has pending flush signals")
+
+		// Reset flush state for future events.
+		app.apmClient.ResetFlush()
```
axw, line 216: Should we call this with a `defer` at the top of the function? In case the flushed request comes, but runtimeDone or the timer also fire.

axw (changes requested):
Almost there :)

Basically I think we should revert the changes to ForwardApmData and EnqueueAPMData, and don't require enqueuing anything to close c.flushCh.

axw (approved):
Thanks for your persistence :smile:
LGTM!

axw:
I've run the Lambda in a loop for a while, and it doesn't appear to panic at all. I did see some unhandled timeout errors, but I get those without the agent or extension enabled too - doesn't appear to be related to the extension.

PR #260, opened 2022-08-01
fix: stop recreating http transports on each info request

kruskall:
http transports should be reused as per documentation. Recreating the
reverse proxy and the transport on each request is also preventing
connection resuse.


axw (approved):
Nice catch!

PR #265, opened 2022-08-02
perf: improve processmetadata performance and reduce allocations

kruskall:
Avoid copying byte slices and make ProcessMetadata zero-allocation.
Add a benchmark to reproduce the numbers:

```
name                       old time/op    new time/op    delta
ProcessMetadata/simple-16    7.89µs ± 8%    0.04µs ± 3%   -99.52%  (p=0.000 n=10+10)

name                       old alloc/op   new alloc/op   delta
ProcessMetadata/simple-16    12.6kB ± 0%     0.0kB       -100.00%  (p=0.000 n=10+10)

name                       old allocs/op  new allocs/op  delta
ProcessMetadata/simple-16      5.00 ± 0%      0.00       -100.00%  (p=0.000 n=10+10)
```

### DIFF
--- a/apm-lambda-extension/apmproxy/metadata.go
+++ b/apm-lambda-extension/apmproxy/metadata.go
@@ -40,12 +40,7 @@ func ProcessMetadata(data AgentData) ([]byte, error) {
 
 	before, _, _ := bytes.Cut(uncompressedData, []byte("\n"))
 
-	// Try without lowercase first to avoid allocations.
 	if bytes.Contains(before, []byte("metadata")) {
-		return before, nil
-	}
-
-	if bytes.Contains(bytes.ToLower(before), []byte("metadata")) {
 		return before, nil
 	}
 
--- a/apm-lambda-extension/apmproxy/metadata.go
+++ b/apm-lambda-extension/apmproxy/metadata.go
@@ -18,14 +18,12 @@
 package apmproxy
 
 import (
-	"bufio"
 	"bytes"
 	"compress/gzip"
 	"compress/zlib"
 	"errors"
 	"fmt"
 	"io"
-	"strings"
 )
 
 type MetadataContainer struct {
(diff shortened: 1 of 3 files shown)


apm-lambda-extension/apmproxy/metadata.go
```diff
@@ -39,18 +37,25 @@ func ProcessMetadata(data AgentData) ([]byte, error) {
...
+	before, _, _ := bytes.Cut(uncompressedData, []byte("\n"))
+
+	// Try without lowercase first to avoid allocations.
+	if bytes.Contains(before, []byte("metadata")) {
+		return before, nil
 	}
+
+	if bytes.Contains(bytes.ToLower(before), []byte("metadata")) {
+		return before, nil
+	}
```
axw, line 50: Maybe we could even get rid of this check? Agents are expected to send a valid `{"metadata":...` line. If they don't, then the agent should be fixed; it's not the Lambda extension's responsibility to filter out junk. WDYT?

axw (commented):
Looks good, just one thought about perhaps getting rid of the `bytes.Contains` checks altogether.

Since `bytes.Cut` was added in Go 1.18, we'll need to update the CI pipeline config to build only with Go 1.18. Not sure why we're testing with multiple versions anyway, since this is not a library -- we should just use one version.

kruskall:
> Since bytes.Cut was added in Go 1.18, we'll need to update the CI pipeline config to build only with Go 1.18. Not sure why we're testing with multiple versions anyway, since this is not a library -- we should just use one version.

Good catch! I updated the job definition. I'll open a new MR to bump the go version (Go 1.19 was also released today)

apm-lambda-extension/apmproxy/metadata.go
```diff
@@ -39,18 +37,20 @@ func ProcessMetadata(data AgentData) ([]byte, error) {
...
-	scanner := bufio.NewScanner(strings.NewReader(string(uncompressedData)))
-	scanner.Scan()
-	if strings.Contains(strings.ToLower(scanner.Text()), "metadata") {
-		return scanner.Bytes(), nil
+
+	before, _, _ := bytes.Cut(uncompressedData, []byte("\n"))
+
+	if bytes.Contains(before, []byte("metadata")) {
+		return before, nil
 	}
```
axw, line 45:
```suggestion
```

To be clear, I don't think we should check it at all, but we should just assume the first line is valid metadata.

axw (approved):
LGTM, just one suggestion on further simplification.

apm-lambda-extension/apmproxy/metadata.go
kruskall: Ops, understood! I've updated the code 👍

Issue #249, opened 2022-07-25
Followup: rework main tests after refactor has been completed


kruskall:
> Did you say you were going to skip the tests until this was done? Looks like the tests in the main package are failing at the moment because of this. If you're going to skip the tests, can you please open an issue so we don't forget to come back to this before creating a new release?

_Originally posted by @axw in https://github.com/elastic/apm-aws-lambda/pull/248#discussion_r928350261_

PR #275, opened 2022-08-09
test: reenable main tests

kruskall:
main_test was disabled after the global state for the logsapi
listener was removed.
The new approach is to choose a random port and pass the address
to the app.
Add a timeout to make sure tests don't hang.

Blocked by https://github.com/elastic/apm-aws-lambda/issues/249

Closes https://github.com/elastic/apm-aws-lambda/issues/249

### DIFF
--- a/apm-lambda-extension/main_test.go
+++ b/apm-lambda-extension/main_test.go
@@ -210,33 +210,50 @@ func newTestStructs(t *testing.T) chan MockEvent {
 func processMockEvent(currId string, event MockEvent, extensionPort string, logsapiAddr string, internals *MockServerInternals, l *zap.SugaredLogger) {
 	sendLogEvent(logsapiAddr, currId, logsapi.Start, l)
 	client := http.Client{}
+
+	// Use a custom transport with a low timeout
+	tr := http.DefaultTransport.(*http.Transport)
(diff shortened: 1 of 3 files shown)


apm-lambda-extension/main_test.go
```diff
@@ -502,191 +540,255 @@ func TestAPMServerRecovery(t *testing.T) {
...
+	}()
+
+	return ctx.Done()
+}
+
+func randomAddr() string {
+	// we cannot return a port that is already in use or it
+	// would return an error when creating the server.
+	// The solution is to spawn a test server to get a random
+	// port and immediately close it so that we can use the port.
```
axw, line 788: This is racy though, as another process could also be trying to acquire an ephemeral port. If the test ends up being flaky,  we may need to have the extension listen on an ephemeral port, log it, and then extract the listening port from the logs. Or something like that.

axw (approved):
Looks great!

apm-lambda-extension/main_test.go
kruskall: Agree, the window to hit the race is small enough that I think it's safe to ignore. Another solution I had in mind was to use unix sockets.

PR #276, opened 2022-08-12
ci: use make in PRs

axw:
Use Make targets, so CI behaves more like developer environments. This ensures license headers are checked in PRs.

Also, tidy up the Makefile a little:
- `zip` now depends on `build`. This ensures a `make zip` in isolation won't zip a stale build and confuse unsuspecting people like me.
- replace `gen-notice` with a non-PHONY `NOTICE.txt` target, by depending on go.mod


kruskall (approved):
Thanks! LGTM!

PR #277, opened 2022-08-12
fix: make zip task compatible with other systems

kruskall:
The zip task was relying on 'touch -d @seconds' to set the timestamp of
output files to allow reproducible builds.
'touch -d @seconds' is not POSIX compliant and the standard only supports
a subset of formats:

https://man7.org/linux/man-pages/man1/touch.1p.html

The standard way of doing it would be to use one of those format or just use
'touch -t TIME'.
Since SOURCE_DATE_EPOCH is providing seconds since epoch and we don't have a
standard way to convert them to a standard format used by touch, we rely on
'date' and use a fallback mechanism to try two different syntax:

- Busybox and GNU coreutils date
- BSD date

### DIFF
--- a/apm-lambda-extension/Makefile
+++ b/apm-lambda-extension/Makefile
@@ -8,6 +8,11 @@ AGENT_VERSION = $(shell echo $${BRANCH_NAME} | cut -f 2 -d 'v')
 # Add support for SOURCE_DATE_EPOCH and reproducble buils
 # See https://reproducible-builds.org/specs/source-date-epoch/
 SOURCE_DATE_EPOCH ?= $(shell git log -1 --pretty=%ct)
+DATE_FMT = +%Y%m%d%H%M.%S
+# Fallback mechanism to support other systems:
+# 1. 'date -d': Busybox and GNU coreutils.
+# 2. 'date -r': BSD date. It does not support '-d'.
+BUILD_DATE = $(shell date -u -d "@${SOURCE_DATE_EPOCH}" "${DATE_FMT}" 2>/dev/null || date -u -r "${SOURCE_DATE_EPOCH}" "${DATE_FMT}")
 
 ifndef GOARCH
 	GOARCH=amd64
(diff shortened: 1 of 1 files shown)


trentm (approved):
Thanks for doing this! WFM on macos, and two runs of `make build zip` result in an identical "extension.zip".

The quote from https://www.gnu.org/software/coreutils/manual/html_node/Date-input-formats.html applies here:

>  It is like a set of trapezoidal building blocks, with no vertical or horizontal surfaces, like a language in which the simplest thought demands ornate constructions, useless particles and lengthy circumlocutions.

working with time is such a pain.

apm-lambda-extension/Makefile
```diff
@@ -8,6 +8,11 @@ AGENT_VERSION = $(shell echo $${BRANCH_NAME} | cut -f 2 -d 'v')
 # Add support for SOURCE_DATE_EPOCH and reproducble buils
 # See https://reproducible-builds.org/specs/source-date-epoch/
 SOURCE_DATE_EPOCH ?= $(shell git log -1 --pretty=%ct)
+DATE_FMT = +%Y%m%d%H%M.%S
+# Fallback mechanism to support other systems:
+# 1. 'date -d': Busybox and GNU coreutils.
+# 2. 'date -r': BSD date. It does not support '-d'.
+BUILD_DATE = $(shell date -u -d "@${SOURCE_DATE_EPOCH}" "${DATE_FMT}" 2>/dev/null || date -u -r "${SOURCE_DATE_EPOCH}" "${DATE_FMT}")
```
axw, line 15: I wonder if it's worth trying to adhere to SOURCE_DATE_EPOCH. This would be simpler by using the `--date` flag of `git log`:

`BUILD_DATE = $(shell git log -1 --pretty=%cd --date=format:%Y%m%d%H%M.%S)`

axw (approved):
LGTM

apm-lambda-extension/Makefile
kruskall: I think we should still support SOURCE_DATE_EPOCH but that's a valid point. 

Moreover avoid calling `date` would increase compatibility since `touch` is POSIX compliant and git is required to build the extension.

I will merge this for now but we can iterate on it!

Issue #261, opened 2022-08-01
Update module name to be consistent with other elastic projects


kruskall:
The current module path is `elastic/apm-lambda-extension` (https://github.com/elastic/apm-aws-lambda/blob/main/apm-lambda-extension/go.mod#L1)

It should be update to follow best practice and be consistent with other elastic projects.

See https://github.com/elastic/apm-agent-go/blob/main/go.mod#L1
See https://github.com/elastic/beats/blob/main/go.mod#L1
See https://github.com/elastic/go-elasticsearch/blob/main/go.mod#L1

PR #280, opened 2022-08-22
fix: update module name to use github url

kruskall:
The previous module name was invalid and using a subfolder to
store the source code was creating additional burden (e.g.
you cannot 'git clone && make'.

Update the module name to use the github url and flatten the
directory structure.

Closes https://github.com/elastic/apm-aws-lambda/issues/261


README.md
```diff
@@ -4,17 +4,86 @@
 
 This repository contains code and tools for working with the Elastic APM AWS Lambda extension.
 
+Elastic APM Agents can be used with AWS Lambda to monitor the execution of your AWS Lambda functions.
+Currently the Node.js, Python and Java Agents are supported.
+
+The extension listens for data once per function invocation and forwards data to an APM Server.
```
axw, line 10: The list of supported agents is in the user docs - what do you think about just dropping these bits?

README.md
```diff
@@ -4,17 +4,86 @@
...
 ## Getting Help
 
 If you find a bug, please [report an issue](https://github.com/elastic/apm-agent-go/issues).
 For any other assistance, please open or add to a topic on the [APM discuss forum](https://discuss.elastic.co/c/apm).
+
+## Contributing
+
+See [contributing](CONTRIBUTING.md) for details about reporting bugs, requesting features, or code contributions.
+
+## Testing unreleased versions
```
axw, line 27: Most of this is developer-oriented, and not relevant to end users. Although we have end-user docs, I think it would still be nice to keep this landing page minimal.

What do you think about moving all of the below with the exception of License and Release to a new TESTING.md doc? License should stay in this doc. And maybe link to RELEASE.md from CONTRIBUTING.md?

axw (commented):
Looks good, just a couple of comments on the readme. Let me know what you think.

README.md
kruskall: Good point! Agree 👍

README.md
kruskall: Good catch! I agree, that's a lot cleaner!

> And maybe link to RELEASE.md from CONTRIBUTING.md?

I'm not sure about this. Release is already linked in the readme and won't provide much value to contributors as it's not something they would worry about. Maybe we should link to TESTING.md ? WDYT ?

TESTING.md
```diff
@@ -1,28 +1,13 @@
...
-See [Monitoring AWS Lambda Functions](https://www.elastic.co/guide/en/apm/guide/current/monitoring-aws-lambda.html) to get started and learn more about available configuration options. 
-
-## Contributing
-
-See [contributing](CONTRIBUTING.md) for details about reporting bugs, requesting features, or code contributions.
+# APM AWS Lambda extension Testing
 
 ## Testing unreleased versions
-If you need to test an unreleased version you can compile the package and it's dependencies and make them available as AWS Lambda Layer. 
+If you need to test an unreleased version you can compile the package and it's dependencies and make them available as AWS Lambda Layer.
```
axw, line 4:
```suggestion
If you need to test an unreleased version you can compile the package and its dependencies and make them available as AWS Lambda Layer.
```

TESTING.md
```diff
@@ -55,7 +40,7 @@ aws lambda publish-layer-version \
  --zip-file  "fileb://extension.zip"
 ```
 
-The out from the above command will include a `LayverVersionArn` field, which contains the unique string identifier for your layer.  The will look something like the following. 
+The out from the above command will include a `LayverVersionArn` field, which contains the unique string identifier for your layer.  The will look something like the following.
```
axw, line 43:
```suggestion
The output from the above command will include a `LayerVersionArn` field, which contains the unique string identifier for your layer.  The will look something like the following.
```

TESTING.md
```diff
@@ -1,28 +1,13 @@
...
 ## Testing unreleased versions
-If you need to test an unreleased version you can compile the package and it's dependencies and make them available as AWS Lambda Layer. 
+If you need to test an unreleased version you can compile the package and it's dependencies and make them available as AWS Lambda Layer.
 
 ### Compile package and dependencies
 
 To run an unreleased version of this extension, you will need to ensure that your build architecture matches that of the Lambda execution environment by compiling with `GOOS=linux` and `GOARCH=amd64` if you are not running in a Linux environment.
 
-To build the extension into the `bin/extensions` folder, run the following commands. 
+To build the extension into the `bin/extensions` folder, run the following commands.
```
axw, line 10:
```suggestion
To build the extension in the `bin/extensions` folder, run the following commands.
```

axw (approved):
Thanks, looks good. Just spotted a few typos in the existing docs.

PR #281, opened 2022-08-22
feat: parse server response and react to error messages

kruskall:
Start backoff mechanism on critical error and log a warning on
client errors.

Any change to the transport status is guarded by a mutex but
some methods were using the field directly, causing race
conditions.

apmproxy tests are using a timeout of 5s to wait for the status
to change during backoff. However there is a subtle bug that
causes the test to fail because the initial delay is exactly 5s.
To avoid this, the tests wait 7s now.

Blocked by https://github.com/elastic/apm-aws-lambda/pull/280

Closes https://github.com/elastic/apm-aws-lambda/issues/225
Closes https://github.com/elastic/apm-aws-lambda/issues/205


apmproxy/apmserver.go
```diff
@@ -131,61 +142,112 @@ func (c *Client) PostToApmServer(ctx context.Context, agentData AgentData) error
...
 	switch status {
 	case Healthy:
 		c.mu.Lock()
 		c.Status = status
 		c.logger.Debugf("APM server Transport status set to %s", c.Status)
 		c.ReconnectionCount = -1
 		c.mu.Unlock()
+	case RateLimited, ClientFailing:
+		// No need to start backoff, this is a temporary status. It usually
+		// means we went over the limit of events/s.
```
axw, line 228: When would ClientFailing be temporary? From what I can see above, it's either due to auth failure (probably not temporary without user intervention?) or some validation error (probably implies a bug in the agent or server?)

apmproxy/apmserver.go
```diff
@@ -30,11 +31,21 @@ import (
 	"time"
 )
 
+type jsonResult struct {
+	Accepted int         `json:"accepted"`
```
axw, line 35: Seeing as we don't use this field, should we remove it from the struct to avoid decoding it?

axw (commented):
Looks good overall, mostly just wondering about the decision not to backoff on auth & validation failures. We should probably have a test or two covering the new states.

apmproxy/apmserver.go
kruskall: Good catch!

apmproxy/apmserver.go
kruskall: > From what I can see above, it's either due to auth failure (probably not temporary without user intervention?)

Ah, good point! I've made auth errors a critical failure

> When would ClientFailing be temporary?

From what I could see from the middlewares in the APM server repository, this would happen on data decoding/validation errors, request body too large or invalid query. Those are errors tied to a specific request, I don't think we should trigger a backoff and associated delay.

apmproxy/apmserver.go
```diff
@@ -131,61 +141,112 @@ func (c *Client) PostToApmServer(ctx context.Context, agentData AgentData) error
...
+	c.logger.Warnf("unhandled status code: %d", resp.StatusCode)
 	return nil
 }
 
-// SetApmServerTransportState takes a state of the APM server transport and updates
+// IsUnhealthy returns true if the apmproxy is not healthy.
+func (c *Client) IsUnhealthy() bool {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return c.Status == Failing
```
simitt, line 207: Given the additiondal locking and that `UpdateStatus` is called on every server response, even if suceeding ([Line 151](https://github.com/elastic/apm-aws-lambda/pull/281/files#diff-646620745ab70102b8587146dbc415a5e6a40c6fe0649830f84ade4862644ab0R151)), have you thought about changing to a `RWMutex` ?

apmproxy/apmserver.go
simitt: > From what I could see from the middlewares in the APM server repository, this would happen on data decoding/validation errors, request body too large or invalid query. Those are errors tied to a specific request, I don't think we should trigger a backoff and associated delay.

good point, I think this is fine as is now.

apmproxy/apmserver.go
```diff
@@ -30,11 +31,20 @@ import (
...
+	Message  string `json:"message"`
+	Document string `json:"document,omitempty"`
+}
+
 // ForwardApmData receives agent data as it comes in and posts it to the APM server.
 // Stop checking for, and sending agent data when the function invocation
 // has completed, signaled via a channel.
 func (c *Client) ForwardApmData(ctx context.Context, metadataContainer *MetadataContainer) error {
-	if c.Status == Failing {
+	if c.IsUnhealthy() {
```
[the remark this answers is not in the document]
kruskall: I don't think so. When the transport state changes to `Failing` a warning is printed out with more details and the backoff mechanism starts. If we log a debug message every time we try to push data while the transport is recovering we might create more noise in the logs.

apmproxy/apmserver.go
kruskall: Great point! I've switched to a `RWMutex` and updated the method to avoid locking when possible. With this approach there should be no lock contention as long as the transport is `Healthy`

apmproxy/apmserver.go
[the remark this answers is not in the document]
simitt: makes sense, thanks

apmproxy/apmserver.go
```diff
@@ -131,61 +141,122 @@ func (c *Client) PostToApmServer(ctx context.Context, agentData AgentData) error
...
 // to the APM server.
 //
 // This function is public for use in tests.
-func (c *Client) SetApmServerTransportState(ctx context.Context, status Status) {
+func (c *Client) UpdateStatus(ctx context.Context, status Status) {
+	// Reduce lock contention as UpdateStatus is called on every
+	// successful request
+	c.mu.RLock()
+	if status == c.Status {
+		c.logger.Debugf("APM server Transport status not changed: %s", status)
```
simitt, line 222: I know it's in debug mode, but that might be overwhelming; agree with a former comment of yours to only lock when status is changed.

apmproxy/apmserver.go
```diff
@@ -131,61 +141,122 @@ func (c *Client) PostToApmServer(ctx context.Context, agentData AgentData) error
...
+		c.mu.RUnlock()
+		return
+	}
+	c.mu.RUnlock()
+
 	switch status {
 	case Healthy:
 		c.mu.Lock()
```
simitt, line 230: Theoretically the status might have changed to `Healthy` between `c.mu.RUnlock()` and `c.mu.Lock()`, I suggest to add another check for `status == c.Status` here before actually changing it.

apmproxy/apmserver.go
kruskall: I agree but I think this is changing the previous behaviour so just to clarify: we should not log a message if the status is not changed, correct ?

apmproxy/apmserver.go
kruskall: Good catch!

apmproxy/apmserver.go
simitt: correct

kruskall: Great! Updated 👍

simitt (approved):
LGTM - thanks for the changes.

simitt:
@axw any other concerns from your side? I believe this is ready to merge otherwise.

axw (approved):
Sorry for the silence, I lost sight of this one. LGTM.

Issue #290, opened 2022-08-25
Disable cgo


axw:
We should build the extension with cgo disabled, to ensure there are no glibc/ABI compatibility issues between the build host and the target Lambda execution environment.

PR #292, opened 2022-08-25
feat: disable cgo

kruskall:
build the extension with cgo disabled, to ensure there are
no glibc/ABI compatibility issues between the build host and
the target Lambda execution environment.

Closes https://github.com/elastic/apm-aws-lambda/issues/290

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -28,6 +28,7 @@ https://github.com/elastic/apm-aws-lambda/compare/v1.1.0...main[View commits]
 
 [float]
 ===== Features
+- Disable CGO to prevent libc/ABI compatibility issues {lambda-pull}292[292]
 
 [float]
 ===== Bug fixes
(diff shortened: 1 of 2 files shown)


axw (approved):
Thanks :)

felixbarny:
Did this have an impact on the size of the binary? If so, what's the before and after? Thanks!

axw:
There's a small reduction in size.

Before:

```
-rwxrwxr-x 1 andrew andrew  11M Aug 24 17:01 bin/extensions/apm-lambda-extension
-rw-rw-r-- 1 andrew andrew 5.5M Aug 26 17:40 bin/extension.zip
```

After:

```
-rwxrwxr-x 1 andrew andrew  10M Aug 26 07:51 bin/extensions/apm-lambda-extension
-rw-rw-r-- 1 andrew andrew 5.4M Aug 26 17:40 bin/extension.zip
```

PR #294, opened 2022-08-29
feat: replace timeout seconds option with timeout durations

kruskall:
Deprecate TIMEOUT_SECONDS options in favour of simpler TIMEOUT options.
The new options take a duration as a more user-friendly and flexible
input. It will always take precedence over the old option.
If the old option is provided, a warning is printed out.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -83,20 +83,18 @@ func New(opts ...configOption) (*App, error) {
 
 	var apmOpts []apmproxy.Option
 
-	receiverTimeoutSeconds, err := getIntFromEnvIfAvailable("ELASTIC_APM_DATA_RECEIVER_TIMEOUT_SECONDS")
-	if err != nil {
-		return nil, err
-	}
-	if receiverTimeoutSeconds != 0 {
-		apmOpts = append(apmOpts, apmproxy.WithReceiverTimeout(time.Duration(receiverTimeoutSeconds)*time.Second))
+	if receiverTimeout, ok, err := parseDurationTimeout(app.logger, "ELASTIC_APM_DATA_RECEIVER_TIMEOUT", "ELASTIC_APM_DATA_RECEIVER_TIMEOUT_SECONDS"); err != nil || ok {
+		if err != nil {
+			return nil, err
+		}
+		apmOpts = append(apmOpts, apmproxy.WithReceiverTimeout(receiverTimeout))
 	}
 
-	dataForwarderTimeoutSeconds, err := getIntFromEnvIfAvailable("ELASTIC_APM_DATA_FORWARDER_TIMEOUT_SECONDS")
-	if err != nil {
-		return nil, err
-	}
-	if dataForwarderTimeoutSeconds != 0 {
-		apmOpts = append(apmOpts, apmproxy.WithDataForwarderTimeout(time.Duration(dataForwarderTimeoutSeconds)*time.Second))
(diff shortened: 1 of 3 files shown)


app/app.go
```diff
@@ -129,17 +127,28 @@ func New(opts ...configOption) (*App, error) {
...
+			return 0, false, fmt.Errorf("failed to parse %s: %w", flag, err)
+		}
+
+		return d, true, nil
 	}
 
-	value, err := strconv.Atoi(strValue)
-	if err != nil {
-		return -1, err
+	l.Warnf("%s is deprecated, please consider moving to %s", deprecatedFlag, flag)
```
axw, line 140: should this only be logged if the deprecated flag is actually specified? i.e. move it to line 143?

docs/monitoring-aws-lambda.asciidoc
```diff
@@ -88,13 +88,19 @@ The configured name of your application or service.  The APM agent will use this
 === `ELASTIC_APM_DATA_RECEIVER_TIMEOUT_SECONDS`
```
axw, line 88: Should we add deprecation notices to the docs for the existing config? https://github.com/elastic/docs#additions-and-deprecations

axw (commented):
Thanks, I like it! Just a couple of minor things.

simitt:
I am curious if you reviewed other configuration options also? 
The lambda extension is GA, and I'd like to avoid spreading deprecations in separate releases. Instead, we should aim for reworking any existing config option that should be changed within the same release.

app/app.go
kruskall: Good catch! Thanks for spotting this, updated!

docs/monitoring-aws-lambda.asciidoc
kruskall: Good point! Added 👍

kruskall:
> I am curious if you reviewed other configuration options also? The lambda extension is GA, and I'd like to avoid spreading deprecations in separate releases. Instead, we should aim for reworking any existing config option that should be changed within the same release.

Good question! I reviewed the other options but these are the only one that needed a rework IMO

docs/monitoring-aws-lambda.asciidoc
```diff
@@ -85,16 +85,36 @@ Instead of specifying the <<aws-lambda-config-authentication-keys>> as plain tex
...
 The configured name of your application or service.  The APM agent will use this value when reporting data to the APM Server. If unset, the APM agent will automatically set the value based on the Lambda function name. Use this config option if you want to group multiple Lambda functions under a single service entity in APM.
 
+[[aws-lambda-config-data-receiver-timeout-seconds]]
 === `ELASTIC_APM_DATA_RECEIVER_TIMEOUT_SECONDS`
+
+deprecated::[1.2.0,Replaced by `ELASTIC_APM_DATA_RECEIVER_TIMEOUT`. See <<aws-lambda-config-data-receiver-timeout>>]
+
 The {apm-lambda-ext}'s timeout value, in seconds, for receiving data from the APM agent. The _default_ is `15`.
 
+=== `ELASTIC_APM_DATA_RECEIVER_TIMEOUT`
```
axw, line 95:
```suggestion
[[aws-lambda-config-data-receiver-timeout]]
=== `ELASTIC_APM_DATA_RECEIVER_TIMEOUT`
```

axw (commented):
I think it looks good, but I can't see the rendered docs as there's a missing anchor.

docs/monitoring-aws-lambda.asciidoc
kruskall: Good catch!

axw (approved):

Issue #217, opened 2022-06-09
Add Smoke Testing to the CI pipeline


jlvoiseux:
Following the deployment of a broken release ([v1.0.1](https://github.com/elastic/apm-aws-lambda/releases/tag/v1.0.1)), it has been suggested to introduce smoke testing to the repository. These smoke tests would take inspiration from [elastic-lambda-extension-testing](https://github.com/elastic/apm-lambda-extension-testing) and would be implemented via Terraform. A possible process outline would be as follows:
- Build the Lambda layer
- Spin up an APM-instrumented Lambda function behind an API Gateway
- Attach the extension layer to the function
- Perform 2 invocations (in order to also check [metrics generation](https://github.com/elastic/apm-aws-lambda/pull/202))
- Check a dedicated ES instance for the presence of target documents

An implementation of SAM CLI-based E2E testing already [exists](https://github.com/elastic/apm-aws-lambda/tree/main/apm-lambda-extension/e2e-testing), but since the SAM CLI does not support the Lambda Logs API used for Metrics, its relevance will decrease by the time we release `v1.1`.

axw:
We might want to take some inspiration from https://github.com/cyrille-leclerc/my-serverless-shopping-cart/tree/main/checkout-function/deploy, which uses https://registry.terraform.io/modules/terraform-aws-modules/lambda/aws/latest

simitt:
@cachedout would someone from your team be able to help out with the CI setup for the newly added smoke tests once https://github.com/elastic/apm-aws-lambda/pull/296 is finalized)?

axw:
Reopening this - we can close it once the CI integration has been done.

cachedout:
> @cachedout would someone from your team be able to help out with the CI setup for the newly added smoke tests once https://github.com/elastic/apm-aws-lambda/pull/296 is finalized)?

@simitt Apologies, this one slipped through my filters. It looks like this has been done by @kruskall in https://github.com/elastic/apm-aws-lambda/pull/310 and that it's currently under review. Is there anything needed beyond that?

simitt:
Nothing more to do here.

PR #296, opened 2022-08-29
feat: add smoke testing

kruskall:
Add a terraform script and an example js function.
The function is pushed to aws with the following layers:
- apm js agent
- locally built lambda extension

Log level and aws region can be configured using variables.

Closes https://github.com/elastic/apm-aws-lambda/issues/217


kruskall:
First iteration, ready for review, any feedback and tip on how to integrate this in CI is really welcomed

tf/main.tf
```diff
@@ -0,0 +1,42 @@
...
+  layers = [
+    module.lambda_layer_local.lambda_layer_arn,
+    "arn:aws:lambda:eu-central-1:267093732750:layer:elastic-apm-node-ver-3-38-0:1",
+  ]
+
+  environment_variables = {
+    NODE_OPTIONS                  = "-r elastic-apm-node/start"
+    ELASTIC_APM_LOG_LEVEL         = var.log_level
+    ELASTIC_APM_LAMBDA_APM_SERVER = "CHANGEME"
+    ELASTIC_APM_SECRET_TOKEN      = "CHANGEME"
```
axw, line 24: Should these be variables?

axw (commented):
This is a great start. Before closing #217 (the description says this PR closes it), we should make invocations to the function and check the results in Elasticsearch.

In standup I mentioned that I think we should use an ephemeral deployment for this. Perhaps for that we can reuse some of the Terraform modules defined in the apm-server repo, like this:

```
module "ec_deployment" {
  source = "github.com/elastic/apm-server//testing/infra/terraform/modules/ec_deployment?depth=1"
  deployment_name_prefix = "apm-aws-lambda"
  integrations_server = true
  apm_server_expvar = false
  apm_server_pprof = false
}
```

(@marclop WDYT about other projects using the apm-server modules like that?)

We could then compose that and the module you have defined to orchestrate the creation of the ESS deployment and Lambda. After provisioning we would run a script that sends requests to the Lambda URL, and then poll Elasticsearch for results.

I've been intending to propose that we take this approach for the agents too, and stop using apm-integration-testing. CC @felixbarny @cachedout

tf/main.tf
```diff
@@ -0,0 +1,42 @@
...
+  function_name = "smoke-testing-test"
+  description   = "Example Lambda function for smoke testing"
+  handler       = "index.handler"
+  runtime       = "nodejs16.x"
+
+  source_path = "../testdata/function/"
+
+  layers = [
+    module.lambda_layer_local.lambda_layer_arn,
+    "arn:aws:lambda:eu-central-1:267093732750:layer:elastic-apm-node-ver-3-38-0:1",
```
axw, line 17:
```suggestion
    "arn:aws:lambda:${var.aws_region}:267093732750:layer:elastic-apm-node-ver-3-38-0:1",
```

tf/main.tf
kruskall: Updated to use the apm server and secret token from the ephemeral deployment

tf/main.tf
kruskall: Good catch!

kruskall:
> In standup I mentioned that I think we should use an ephemeral deployment for this. Perhaps for that we can reuse some of the Terraform modules defined in the apm-server repo, like this:
> 
> module "ec_deployment" {
>   source = "github.com/elastic/apm-server//testing/infra/terraform/modules/ec_deployment?depth=1"
>   deployment_name_prefix = "apm-aws-lambda"
>   integrations_server = true
>   apm_server_expvar = false
>   apm_server_pprof = false
> }
> 
> We could then compose that and the module you have defined to orchestrate the creation of the ESS deployment and Lambda. After provisioning we would run a script that sends requests to the Lambda URL, and then poll Elasticsearch for results.

Could you clarify this ? Looking at the module in the apm-server repository I don't think we will end up using much of the available inputs. Is there a specific feature we're planning to use ?

axw:
> Could you clarify this ? Looking at the module in the apm-server repository I don't think we will end up using much of the available inputs. Is there a specific feature we're planning to use ?

My initial thought was also to use the `ec` provider directly, which I assume is what you're thinking here. The main thing we would get from using `ec_deployment` from the apm-server repo is extraction of the generated secret token from the APM integration policy.

tf/variables.tf
```diff
@@ -0,0 +1,23 @@
...
+  default     = "trace"
+}
+
+variable "ec_region" {
+  type        = string
+  description = "ec region"
+  default     = "aws-eu-central-1"
+}
+
+variable "ec_deployment_template" {
```
axw, line 19: Do we need this? Maybe start without it?

tf/main.tf
```diff
@@ -0,0 +1,89 @@
...
+    topology {
+      id         = "hot_content"
+      size       = "1g"
+      zone_count = 1
+    }
+  }
+
+  kibana {}
+
+  apm {}
```
axw, line 45: This is legacy, and will eventually be removed, so we should be using "integrations_server" here instead. The problem then becomes how to fetch the generated secret token - hence ec_deployment.

tf/variables.tf
```diff
@@ -0,0 +1,23 @@
...
+  default     = "eu-central-1"
+}
+
+variable "log_level" {
+  type        = string
+  description = "lambda extension log level"
+  default     = "trace"
+}
+
+variable "ec_region" {
```
axw, line 13:
```suggestion
variable "ess_region" {
```

For consistency with apm-server modules testing/cloud, testing/benchmark, and testing/rally-cloud.

tf/main.tf
```diff
@@ -0,0 +1,89 @@
...
+  elasticsearch {
+    autoscale = "false"
+
+    topology {
+      id         = "hot_content"
+      size       = "1g"
+      zone_count = 1
+    }
+  }
+}
```
axw, line 27: I think it's overkill to create a separate observability deployment for smoke tests. In the future we'll be able to point a deployment at itself for observability, if we need this: https://github.com/elastic/terraform-provider-ec/pull/502 (not yet released)

axw (changes requested):
Looking good @kruskall, just a few changes I'd like to see.

In addition to the comments I left, can we:
 - add .gitignore entries for terraform state?
 - add outputs for the Lambda function URL, and ESS deployment details (Elasticsearch + Kibana URLs, username and password)?

Perhaps we should default the AWS region to gcp-us-west-2, since it's where we do the rest of our cloud-based testing. Although it's crossing regions, the amount of traffic will be minimal - and running the deployment in the dedicated CFT us-west-2 region means we won't be contending with customers for ESS resources.

How will we invoke the Lambda function from some test script after provisioning? Should we be configuring a function URL, and emitting it in output?

cachedout:
> I've been intending to propose that we take this approach for the agents too, and stop using apm-integration-testing. CC @felixbarny @cachedout

This topic has come up a few times lately.

@axw How would you like to proceed in terms of making a decision on a path forward?

axw:
@cachedout I've just created a proposal (in a private repo for reasons, but the gist is the same as what I've described above), I'll ping you on it.

tf/main.tf
kruskall: Agree, should we wait for the next release of the terraform provider or pitch someone about it ?

axw: I would suggest we just remove it for now, and consider adding it back later when "self" is an option. I don't think it's needed for smoke testing, is it?

tf/variables.tf
kruskall: This is required to fully support custom region. Different regions have different templates (there are some regions with common templates however)

tf/main.tf
kruskall: Solved by switching to APM server custom module

tf/variables.tf
kruskall: Good point!

tf/main.tf
kruskall: Solved by switching to custom APM server module!

tf/variables.tf
```diff
@@ -0,0 +1,23 @@
...
+variable "log_level" {
+  type        = string
+  description = "lambda extension log level"
+  default     = "trace"
+}
+
+variable "ess_region" {
+  type        = string
+  description = "ess region"
+  default     = "aws-eu-central-1"
```
axw, line 16: Should we default `ess_region` to `gcp-us-west2` (CFT/Cloud First Testing region), since it's where we do the rest of our cloud-based testing?

Although it's crossing regions, the amount of traffic will be minimal - and running the deployment in the dedicated CFT region means we won't be contending with customers for ESS resources.

tf/variables.tf
```diff
@@ -0,0 +1,23 @@
...
+}
+
+variable "ec_deployment_template" {
```
axw, line 19:
```suggestion
variable "ess_deployment_template" {
```

tf/variables.tf
```diff
@@ -0,0 +1,23 @@
...
+  type        = string
+  description = "ec deployment template"
```
axw, line 21:
```suggestion
  description = "Elastic Cloud deployment template"
```

axw (changes requested):
Looking good, thanks for the updates @kruskall.

I think we should change the default ESS region, but let me know if you have any concerns about that. Ideally I think we would use the defaults for `ess_region` and `deployment_template` as defined in `ec_deployment`, to keep things defined in one place.

In order to close https://github.com/elastic/apm-aws-lambda/issues/217 we will need to actually run some invocations and verify trace events end up in Elasticsearch as expected. I suggest we get this PR finished up and follow up with a test script that does those bits. Can you please remove the "Closes" from the description?

tf/variables.tf
kruskall: Agree! Moved 👍

kruskall:
Ok, I added the script to verify the traces are in Elasticsearch, I think this is ready for another review

tf/test.sh
```diff
@@ -0,0 +1,48 @@
...
+
+trap "cleanup" EXIT
+
+echo "-> Creating the underlying infrastructure..."
+terraform init | tee tf.log
+terraform apply -auto-approve | tee -a tf.log
+
+echo "-> Calling the lambda function..."
+aws lambda invoke --function-name smoke-testing-test response.json
+aws lambda invoke --function-name smoke-testing-test response.json
```
axw, line 20:
```suggestion
aws lambda invoke --region=eu-central-1 --function-name smoke-testing-test response.json
aws lambda invoke --region=eu-central-1 --function-name smoke-testing-test response.json
```

If you don't do this, the CLI's configured default region is used.

axw (approved):
Thanks, this is great. LGTM with the region explicitly passed to the AWS CLI.

marclop (commented):
oops, my review didn't make it pre-merge, I guess we can see if it's worth as a follow up

tf/main.tf
```diff
@@ -0,0 +1,52 @@
...
+
+module "ec_deployment" {
+  source                 = "github.com/elastic/apm-server//testing/infra/terraform/modules/ec_deployment?depth=1"
+  deployment_name_prefix = "apm-aws-lambda-smoke-testing"
+  integrations_server    = true
+  apm_server_expvar      = false
+  apm_server_pprof       = false
+  region                 = var.ess_region
+  deployment_template    = var.ess_deployment_template
```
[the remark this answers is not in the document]
axw: ++ I've opened https://github.com/elastic/apm-aws-lambda/issues/307 to track this

Issue #256, opened 2022-07-29
Collect Lambda Logs (POC)


simitt:
The AWS Lambda extension currently gathers traces and Lambda metrics. This issue is about extending the current extension to also support collecting log events out of the box. 

The Lambda runtime provides the Lambda function / application logs through the Lambda Logs API directly to the extension. Currently the extension subscribes to the _Logs API_ to retrieve `platform` type logs for collecting metrics. 

To also collect application logs, the extension needs to also subscribe to `function` type logs:
- Receive the logs through the Lambda Logs API
- Enrich the logs with meta information (e.g. `faas.*` fields, `cloud.*` fields, etc.)
- Send the logs to the APM server's new logs intake (https://github.com/elastic/apm-server/issues/8757)

## Scope
* [ ] The collected lambda logs should be available in the APM UI's logs tab for the APM service that represents the Lambda function. 
* [ ] The logs are enriched with meta information. 
* [ ] Preliminary performance tests show that the tradeoff between shipping the logs directly through the extension and the implied latency overhead is reasonable.

This can be treated as a `meta` issue and be split up into further implementation tasks if necessary.

PR #303, opened 2022-09-13
Add support for collecting and sending function logs to APM Server

lahsivjar:
Adds support for collecting `function` logs using AWS Logs API and shipping it to APM-Server with collected agent metadata.

Closes: #256


logsapi/functionlogs.go
```diff
@@ -0,0 +1,99 @@
...
+	Log *logLine
+}
+
+type logLine struct {
+	Timestamp model.Time
+	Message   string
+	FAAS      *model.FAAS
+}
+
+func (l *logLine) MarshalFastJSON(w *fastjson.Writer) error {
```
felixbarny, line 37: Just an idea that could reduce allocations: instead of converting `extension.NextEventResponse` and `LogEvent` into a `LogContainer`, you could pass in `extension.NextEventResponse` and `LogEvent` directly to the `MarshalFastJSON` method and avoid creating these temporary serialization objects.

However, if you feel like the code is cleaner this way and if this isn't contributing a lot to overall allocations, feel free to ignore.

lahsivjar: I ran some benchmarks on the idea; below are the results:

```
# With `LogContainer` and `logLine`
BenchmarkProcessFunctionLog-10    	 1758667	       684.3 ns/op	    1264 B/op	       7 allocs/op

# Without `LogContainer` and `logLine`
BenchmarkProcessFunctionLog-10    	 1808710	       668.9 ns/op	    1184 B/op	       6 allocs/op
```

Based on the results I am a bit more inclined toward keeping the objects since they will allow for easier extension in case more fields need to be added.

WDYT given the benchmark results?

<details>
<summary>(Code diff that produced the 2nd benchmark)</summary>

```diff
diff --cc logsapi/functionlogs.go
index 984b15f,3213428..0000000
--- a/logsapi/functionlogs.go
+++ b/logsapi/functionlogs.go
@@@ -24,28 -24,61 +24,23 @@@ import 
  	"go.elastic.co/fastjson"
  )
  
- type logContainer struct {
 -type LogContainer struct {
--	Log *logLine
--}
--
--type logLine struct {
--	Timestamp model.Time
--	Message   string
--	FAAS      *model.FAAS
--}
--
--func (l *logLine) MarshalFastJSON(w *fastjson.Writer) error {
 -	var firstErr error
 -	w.RawByte('{')
 -	w.RawString("\"message\":")
 -	w.String(l.Message)
 -	w.RawString(",\"@timestamp\":")
 -	if err := l.Timestamp.MarshalFastJSON(w); err != nil && firstErr == nil {
 -		firstErr = err
 -	}
 -	if l.FAAS != nil {
 -		w.RawString(",\"faas\":")
 -		if err := l.FAAS.MarshalFastJSON(w); err != nil && firstErr == nil {
 -			firstErr = err
 -		}
 -	}
 -	w.RawByte('}')
 -	return firstErr
 -}
 -
 -func (lc LogContainer) MarshalFastJSON(json *fastjson.Writer) error {
 -	json.RawString(`{"log":`)
 -	if err := lc.Log.MarshalFastJSON(json); err != nil {
 -		return err
 -	}
 -	json.RawString(`}`)
 -	return nil
 -}
 -
+ func marshalFastJSON(w *fastjson.Writer, fd *extension.NextEventResponse, log LogEvent) error {
 -	var firstErr error // 4
 +	var firstErr error
+ 	w.RawString(`{"log":`)
  	w.RawByte('{')
  	w.RawString("\"message\":")
- 	w.String(l.Message)
+ 	w.String(log.StringRecord)
  	w.RawString(",\"@timestamp\":")
- 	if err := l.Timestamp.MarshalFastJSON(w); err != nil && firstErr == nil {
+ 	if err := model.Time(log.Time).MarshalFastJSON(w); err != nil && firstErr == nil {
  		firstErr = err
  	}
- 	if l.FAAS != nil {
+ 	if fd != nil {
  		w.RawString(",\"faas\":")
- 		if err := l.FAAS.MarshalFastJSON(w); err != nil && firstErr == nil {
+ 		faas := model.FAAS{
+ 			Execution: fd.RequestID,
+ 			ID:        fd.InvokedFunctionArn,
+ 		}
+ 		if err := faas.MarshalFastJSON(w); err != nil && firstErr == nil {
  			firstErr = err
  		}
  	}
@@@ -69,23 -94,8 +56,8 @@@ func ProcessFunctionLog
  	functionData *extension.NextEventResponse,
  	log LogEvent,
  ) (apmproxy.AgentData, error) {
- 	lc := logContainer{
- 		Log: &logLine{
- 			Timestamp: model.Time(log.Time),
- 			Message:   log.StringRecord,
- 		},
- 	}
- 
- 	if functionData != nil {
- 		// FaaS Fields
- 		lc.Log.FAAS = &model.FAAS{
- 			Execution: functionData.RequestID,
- 			ID:        functionData.InvokedFunctionArn,
- 		}
- 	}
- 
 -	var jsonWriter fastjson.Writer // 1
 +	var jsonWriter fastjson.Writer
- 	if err := lc.MarshalFastJSON(&jsonWriter); err != nil {
+ 	if err := marshalFastJSON(&jsonWriter, functionData, log); err != nil {
  		return apmproxy.AgentData{}, err
  	}
  
diff --git a/logsapi/functionlogs_test.go b/logsapi/functionlogs_test.go
index 64d4312..5aed6a6 100644
--- a/logsapi/functionlogs_test.go
+++ b/logsapi/functionlogs_test.go
@@ -68,3 +68,23 @@ func TestProcessFunctionLog(t *testing.T) {
 		)
 	})
 }
+
+func BenchmarkProcessFunctionLog(b *testing.B) {
+	mc := &apmproxy.MetadataContainer{
+		Metadata: []byte(`{"metadata":{"service":{"agent":{"name":"apm-lambda-extension","version":"1.1.0"},"framework":{"name":"AWS Lambda","version":""},"language":{"name":"python","version":"3.9.8"},"runtime":{"name":"","version":""},"node":{}},"user":{},"process":{"pid":0},"system":{"container":{"id":""},"kubernetes":{"node":{},"pod":{}}},"cloud":{"provider":"","instance":{},"machine":{},"account":{},"project":{},"service":{}}}}`),
+	}
+	event := LogEvent{
+		Time:         time.Date(2022, 11, 12, 0, 0, 0, 0, time.UTC),
+		Type:         FunctionLog,
+		StringRecord: "ERROR encountered. Stack trace:\nmy-function (line 10)\n",
+	}
+	nextEventResp := &extension.NextEventResponse{
+		RequestID:          "8476a536-e9f4-11e8-9739-2dfe598c3fcd",
+		InvokedFunctionArn: "arn:aws:lambda:us-east-2:123456789012:function:custom-runtime",
+	}
+
+	for n := 0; n < b.N; n++ {
+		_, err := ProcessFunctionLog(mc, nextEventResp, event)
+		require.NoError(b, err)
+	}
+}
```
</details>

felixbarny: Seems to not make a huge difference allocation-wise.

> Based on the results I am a bit more inclined toward keeping the objects since they will allow for easier extension in case more fields need to be added.

SGTM

logsapi/event.go
```diff
@@ -19,38 +19,44 @@ package logsapi
...
-	"github.com/elastic/apm-aws-lambda/extension"
-	"github.com/elastic/apm-aws-lambda/apmproxy"
 	"time"
+
+	"github.com/elastic/apm-aws-lambda/apmproxy"
+	"github.com/elastic/apm-aws-lambda/extension"
 )
 
-// EventType represents the type of logs in Lambda
-type EventType string
```
lahsivjar, line 28: [**For reviewers**]: I have renamed this from `EventType` -> `LogType` to avoid confusion with the `EventType` used for lifecycle events. Let me know if it doesn't sound correct.

logsapi/event.go
```diff
@@ -101,6 +108,18 @@ func (lc *Client) ProcessLogs(
 					lc.logger.Warn("report event request id didn't match the previous event id")
 					lc.logger.Debug("Log API runtimeDone event request id didn't match")
 				}
+			case FunctionLog:
+				if metadataContainer != nil && len(metadataContainer.Metadata) > 0 {
```
lahsivjar, line 112: [**For reviewers**]: The way we collect metadata currently allows function logs to be emitted before metadata can be collected. Since APM Server considers any request without metadata to be `400` the PR drops any log lines before metadata is collected. Another approach can be to buffer these log lines and send them in batch but that might increase the resource utilization of the extension. Let me know if there are any concerns/alternative approaches here.

logsapi/functionlogs_test.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+	event := LogEvent{
+		Time:         time.Date(2022, 11, 12, 0, 0, 0, 0, time.UTC),
+		Type:         FunctionLog,
+		StringRecord: "ERROR encountered. Stack trace:my-function (line 10)",
+	}
+	reqID := "8476a536-e9f4-11e8-9739-2dfe598c3fcd"
+	invokedFnArn := "arn:aws:lambda:us-east-2:123456789012:function:custom-runtime"
+	expectedData := fmt.Sprintf(
+		"%s\n{\"log\":{\"message\":\"%s\",\"@timestamp\":%d,\"faas\":{\"coldstart\":false,\"execution\":\"%s\",\"id\":\"%s\"}}}",
```
lahsivjar, line 42: [**For reviewers**] `coldstart` value here might not be correct and (CMIIW) I don't think there is a way for us to know about this by using `function` log event type. However, the default marshal logic always adds `coldstart` field.

I am thinking it might be better for us to have a custom unmarshal for `FAAS` here. Happy to discuss other thoughts on this.

lahsivjar: While testing the feature, I noticed that it is possible to get `function` log events while a different request ID is being processed. The payload of `function` log doesn't include any `requestID` for correlation with other events.

Based on current knowledge, I think we will have to drop `FAAS` fields from the log body completely.

lahsivjar: For posterity, discussed this with @axw and concluded with the following:

1. Add a custom Marshal logic for `faas`
2. Use `platform.start` and `platform.end` events to track request ID for function logs (assuming function logs will always be bounded between these two events)

felixbarny: that sounds like a smart solution!

lahsivjar: This is fixed now.

logsapi/event.go
```diff
@@ -101,6 +99,23 @@ func (lc *Client) ProcessLogs(
...
```
axw, line 103: Could we neaten this a little by doing this?

```go
if metadataContainer == nil || len(metadataContainer.Metadata) == 0 {
    lc.logger.Warn("Function log received before metadata populated, quietly dropping log")
    continue
}

lc.logger.Debug("Received function log")
...
```

logsapi/functionlogs.go
```diff
@@ -0,0 +1,118 @@
...
+	return firstErr
+}
+
+// faas struct is a subset of go.elastic.co/apm/v2/model#FAAS
+//
+// The purpose of having a separate struct is to have a custom
+// marshalling logic that is targeted for the faas fields
+// available for function logs. For example: `coldstart` value
+// cannot be inferred for function logs so this struct drops
+// the field entirely.
```
axw, line 60: Doesn't the extension get initialised once per cold start? Meaning that we could set `coldstart: true` for the first invocation processed by the extension, and `coldstart: false` for every request thereafter?

We don't need to change this now, but I think we _could_ improve it in the future. We probably don't need to set this property on every log message anyway, though.

lahsivjar: > Doesn't the extension get initialised once per cold start? Meaning that we could set coldstart: true for the first invocation processed by the extension, and coldstart: false for every request thereafter?

Hmm, that's an amazing idea. I will create an issue for this.

main.go
```diff
@@ -39,12 +40,19 @@ func main() {
...
-	app, err := app.New(ctx,
+	appConfigs := []app.ConfigOption{
 		app.WithExtensionName(filepath.Base(os.Args[0])),
 		app.WithLambdaRuntimeAPI(os.Getenv("AWS_LAMBDA_RUNTIME_API")),
 		app.WithLogLevel(os.Getenv("ELASTIC_APM_LOG_LEVEL")),
 		app.WithAWSConfig(cfg),
-	)
+	}
+
+	disableFnLog, _ := strconv.ParseBool(os.Getenv("ELASTIC_DISABLE_FUNCTION_LOG_SUBSCRIPTION"))
```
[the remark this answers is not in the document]
lahsivjar: Initially, I kept a negative flag to keep it `false` by default since not providing the config seemed more aligned with a `false` value. Agree with double-negs being confusing to users, I have updated it to use `ELASTIC_APM_LAMBDA_CAPTURE_LOGS`.

logsapi/event.go
lahsivjar: I have moved the nil handling inside `ProcessFunctionLog`, felt it would be a bit cleaner this way. Let me know your thoughts.

axw: Looks good, thanks!

lahsivjar:
@axw Apologies for the force push 🙇 

[This is the commit](https://github.com/elastic/apm-aws-lambda/pull/303/commits/7e6865cb61e2b52e29f09cc2944ee186bfeaa7e0) added for addressing the review comments.

axw (approved):
Looks great, thanks for the updates :)

logsapi/functionlogs.go
```diff
@@ -0,0 +1,118 @@
...
+	json.RawString(`{"log":`)
+	if err := lc.Log.MarshalFastJSON(json); err != nil {
+		return err
+	}
+	json.RawByte('}')
+	return nil
+}
+
+// ProcessFunctionLog consumes extension event, agent metadata and log
+// event from Lambda logs API to create a payload for APM server
```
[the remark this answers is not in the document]
lahsivjar: Created an issue: https://github.com/elastic/apm-aws-lambda/issues/311

kruskall:
Tested with `BC2` (and `8.6.0-SNAPSHOT`) by provisioning the environment with the terraform script (`SKIP_DESTROY=1` has been used to inspect the cluster) with `TF_VAR_ess_version=8.5.0-SNAPSHOT`.

Verified that the logs are visible in the logs tab in the Observability section.

Made sure that `ELASTIC_APM_LAMBDA_CAPTURE_LOGS=false` disables logs collection.

Possible concern: https://github.com/elastic/apm-aws-lambda/issues/322
I don't think the above issue is a blocker according to the definition of a blocker so I'm going to label this PR as test-plan-ok

RyanHealey1307:
I'm aware this is still experimental, but some feedback...
I've tested this for a python lambda, using the latest APM layer, and when you invocate the function, the logs don't go through to APM until the next invocation. Eg, if my lambda is run at 10:09 and then again at 10:13, the logs from 10:09 are not visible until next next invocation at 10:13. My timeout is set to 5 mins and it's a very simple function, so I can't be a timeout issue.

PR #314, opened 2022-09-22
Batch lambda logs API data before sending to APM-Server

### LINKED ISSUE #311: Batch data collected by log streams before sending to APM-Server
**Is your feature request related to a problem? Please describe.**
Currently, the extension will drop logevents from subscribed logstreams if they are received before agent metadata is available. 

**Describe the solution you'd like**
Extension should buffer data after processing and should send it only after metadata is available. Additionally, extension should also batch data to be more efficient.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

lahsivjar:
### Motivation
The PR is motivated by 2 factors:

1. We send data collected from logs API one by one as they are received. This PR allows batching of the data before sending it to APM-Server. The batches are explicitly sized based on the number of events as well as the age of the batch based on the first entry added to it.
2. Before this PR, the extension drops data until metadata from agents is available. This PR adds logic to buffer events in the extension and if the buffer is full, pushback on Lambda Logs API until metadata is available. Due to this behavior, the lifecycle of the extension might also be affected since we push back on all log events from the Logs API including the `platform.runtimeDone`. In case lambda Logs API drops logs due to resource constraints generated by our pushback, a new event `platform.logsDropped` is published which we log from the extension. The retention of this log event by the logs API is not clear, it might be possible that this log event is also dropped but it hasn't occurred in any of my tests so far. However, assuming that metadata is available without much delay this might not be a deal-breaker.

### Related Issues
Closes #311

### DIFF
--- a/apmproxy/apmserver.go
+++ b/apmproxy/apmserver.go
@@ -324,6 +324,8 @@ func (c *Client) forwardAPMDataByType(ctx context.Context, apmData APMData, batc
 				return fmt.Errorf("failed to extract metadata from agent payload %w", err)
 			}
 			c.setMetadata(metadata)
+			// broadcast that metadata is available
+			close(c.metadataAvailable)
 		}
 		return c.PostToApmServer(ctx, apmData)
 	case Lambda:
@@ -340,10 +342,8 @@ func (c *Client) forwardAPMDataByType(ctx context.Context, apmData APMData, batc
 }
 
 func (c *Client) sendBatch(ctx context.Context, batch *BatchData) error {
-	// TODO: @lahsivjar keep buffering logs in DataChannel until metadata
-	// is available.
 	if len(c.metadata) == 0 {
-		return errors.New("metadata not yet populated, dropping data")
+		return errors.New("unexpected state, metadata should always be populated")
 	}
 	if batch.Size() == 0 {
 		return nil
--- a/apmproxy/apmserver.go
+++ b/apmproxy/apmserver.go
@@ -48,20 +48,7 @@ func (c *Client) ForwardApmData(ctx context.Context) error {
 		return nil
 	}
 
-	// Wait for metadata to be available, metadata will be available as soon as
-	// the first agent data is processed.
-	select {
-	case <-ctx.Done():
-		c.logger.Debug("Invocation context cancelled, not processing any more agent data")
-		return nil
-	case data := <-c.AgentDataChannel:
-		if err := c.forwardAgentData(ctx, data); err != nil {
-			return err
-		}
-	case <-c.metadataAvailable:
-	}
-
-	// Process lambda and agent data after metadata is available
+	var lambdaDataChan chan APMData
 	for {
 		select {
 		case <-ctx.Done():
@@ -71,7 +58,10 @@ func (c *Client) ForwardApmData(ctx context.Context) error {
 			if err := c.forwardAgentData(ctx, data); err != nil {
 				return err
 			}
-		case data := <-c.LambdaDataChannel:
+			// Wait for metadata to be available, metadata will be available as soon as
+			// the first agent data is processed.
+			lambdaDataChan = c.LambdaDataChannel
+		case data := <-lambdaDataChan:
 			if err := c.forwardLambdaData(ctx, data); err != nil {
 				return err
 			}
@@ -88,26 +78,15 @@ func (c *Client) FlushAPMData(ctx context.Context) {
 	c.logger.Debug("Flush started - Checking for agent data")
 
 	// Flush agent data first to make sure metadata is available if possible
-	func() {
-		for {
-			select {
-			case <-ctx.Done():
-				c.logger.Debugf("Failed to flush completely, may result in data drop")
-			case data := <-c.AgentDataChannel:
-				if err := c.forwardAgentData(ctx, data); err != nil {
-					c.logger.Errorf("Error sending to APM Server, skipping: %v", err)
-				}
-			default:
-				c.logger.Debug("Flush ended for agent data - no data in buffer")
-				return
-			}
+	for i := len(c.AgentDataChannel); i > 0; i-- {
+		data := <-c.AgentDataChannel
+		if err := c.forwardAgentData(ctx, data); err != nil {
+			c.logger.Errorf("Error sending to APM Server, skipping: %v", err)
 		}
-	}()
+	}
 
 	// If metadata still not available then fail fast
-	select {
-	case <-c.metadataAvailable:
-	default:
+	if c.batch == nil {
 		c.logger.Warnf("Metadata not available at flush, skipping sending lambda data to APM Server")
 		return
 	}
@@ -355,8 +334,6 @@ func (c *Client) forwardAgentData(ctx context.Context, apmData APMData) error {
 			return fmt.Errorf("failed to extract metadata from agent payload %w", err)
 		}
 		c.batch = NewBatch(c.maxBatchSize, c.maxBatchAge, metadata)
-		// broadcast that metadata is available
-		close(c.metadataAvailable)
 	}
 	return c.PostToApmServer(ctx, apmData)
 }
--- a/apmproxy/apmserver.go
+++ b/apmproxy/apmserver.go
@@ -94,13 +94,14 @@ func (c *Client) FlushAPMData(ctx context.Context) {
 	// Flush lambda data
 	for {
 		select {
-		case <-ctx.Done():
-			c.logger.Debugf("Failed to flush completely, may result in data drop")
 		case apmData := <-c.LambdaDataChannel:
 			c.logger.Debug("Flush in progress - Processing lambda data")
 			if err := c.forwardLambdaData(ctx, apmData); err != nil {
 				c.logger.Errorf("Error sending to APM server, skipping: %v", err)
 			}
+		case <-ctx.Done():
+			c.logger.Debugf("Failed to flush completely, may result in data drop")
+			return
 		default:
 			// Flush any remaining data in batch
 			if err := c.sendBatch(ctx); err != nil {
--- a/apmproxy/apmserver.go
+++ b/apmproxy/apmserver.go
@@ -55,7 +55,7 @@ func (c *Client) ForwardApmData(ctx context.Context) error {
 		c.logger.Debug("Invocation context cancelled, not processing any more agent data")
 		return nil
 	case data := <-c.AgentDataChannel:
-		if err := c.forwardAPMDataByType(ctx, data); err != nil {
+		if err := c.forwardAgentData(ctx, data); err != nil {
 			return err
 		}
 	case <-c.metadataAvailable:
@@ -68,11 +68,11 @@ func (c *Client) ForwardApmData(ctx context.Context) error {
 			c.logger.Debug("Invocation context cancelled, not processing any more agent data")
 			return nil
 		case data := <-c.AgentDataChannel:
-			if err := c.forwardAPMDataByType(ctx, data); err != nil {
+			if err := c.forwardAgentData(ctx, data); err != nil {
 				return err
 			}
 		case data := <-c.LambdaDataChannel:
-			if err := c.forwardAPMDataByType(ctx, data); err != nil {
+			if err := c.forwardLambdaData(ctx, data); err != nil {
 				return err
 			}
 		}
@@ -94,7 +94,7 @@ func (c *Client) FlushAPMData(ctx context.Context) {
 			case <-ctx.Done():
 				c.logger.Debugf("Failed to flush completely, may result in data drop")
 			case data := <-c.AgentDataChannel:
-				if err := c.forwardAPMDataByType(ctx, data); err != nil {
+				if err := c.forwardAgentData(ctx, data); err != nil {
 					c.logger.Errorf("Error sending to APM Server, skipping: %v", err)
 				}
 			default:
@@ -119,7 +119,7 @@ func (c *Client) FlushAPMData(ctx context.Context) {
 			c.logger.Debugf("Failed to flush completely, may result in data drop")
 		case apmData := <-c.LambdaDataChannel:
 			c.logger.Debug("Flush in progress - Processing lambda data")
-			if err := c.forwardAPMDataByType(ctx, apmData); err != nil {
+			if err := c.forwardLambdaData(ctx, apmData); err != nil {
 				c.logger.Errorf("Error sending to APM server, skipping: %v", err)
 			}
 		default:
@@ -348,35 +348,32 @@ func (c *Client) WaitForFlush() <-chan struct{} {
 	return c.flushCh
 }
 
-func (c *Client) forwardAPMDataByType(ctx context.Context, apmData APMData) error {
-	switch apmData.Type {
-	case Agent:
-		if c.batch == nil {
-			metadata, err := ProcessMetadata(apmData)
-			if err != nil {
-				return fmt.Errorf("failed to extract metadata from agent payload %w", err)
-			}
-			c.batch = NewBatch(c.maxBatchSize, c.maxBatchAge, metadata)
-			// broadcast that metadata is available
-			close(c.metadataAvailable)
+func (c *Client) forwardAgentData(ctx context.Context, apmData APMData) error {
+	if c.batch == nil {
+		metadata, err := ProcessMetadata(apmData)
+		if err != nil {
+			return fmt.Errorf("failed to extract metadata from agent payload %w", err)
 		}
-		return c.PostToApmServer(ctx, apmData)
-	case Lambda:
-		if c.batch == nil {
-			// This state is not possible since we are pushing back on
-			// lambda logs API until metadata is available.
-			return errors.New("unexpected state, metadata not yet set")
-		}
-		if err := c.batch.Add(apmData); err != nil {
-			c.logger.Warnf("Dropping data due to error: %v", err)
-		}
-		if c.batch.ShouldShip() {
-			return c.sendBatch(ctx)
-		}
-		return nil
-	default:
-		return errors.New("invalid apm data type")
+		c.batch = NewBatch(c.maxBatchSize, c.maxBatchAge, metadata)
+		// broadcast that metadata is available
+		close(c.metadataAvailable)
 	}
+	return c.PostToApmServer(ctx, apmData)
+}
+
+func (c *Client) forwardLambdaData(ctx context.Context, apmData APMData) error {
+	if c.batch == nil {
+		// This state is not possible since we are pushing back on
+		// lambda logs API until metadata is available.
+		return errors.New("unexpected state, metadata not yet set")
+	}
+	if err := c.batch.Add(apmData); err != nil {
+		c.logger.Warnf("Dropping data due to error: %v", err)
+	}
+	if c.batch.ShouldShip() {
+		return c.sendBatch(ctx)
+	}
+	return nil
 }
 
 func (c *Client) sendBatch(ctx context.Context) error {
--- a/apmproxy/apmserver.go
+++ b/apmproxy/apmserver.go
@@ -93,7 +93,7 @@ func (c *Client) FlushAPMData(ctx context.Context) {
 // The function compresses the APM agent data, if it's not already compressed.
 // It sets the APM transport status to failing upon errors, as part of the backoff
 // strategy.
-func (c *Client) PostToApmServer(ctx context.Context, agentData AgentData) error {
+func (c *Client) PostToApmServer(ctx context.Context, agentData APMData) error {
 	// todo: can this be a streaming or streaming style call that keeps the
 	//       connection open across invocations?
 	if c.IsUnhealthy() {
@@ -283,7 +283,7 @@ func (c *Client) ComputeGracePeriod() time.Duration {
 
 // EnqueueAPMData adds a AgentData struct to the agent data channel, effectively queueing for a send
 // to the APM server.
-func (c *Client) EnqueueAPMData(agentData AgentData) {
+func (c *Client) EnqueueAPMData(agentData APMData) {
 	select {
 	case c.DataChannel <- agentData:
 		c.logger.Debug("Adding agent data to buffer to be sent to apm server")
--- a/apmproxy/apmserver.go
+++ b/apmproxy/apmserver.go
@@ -47,15 +47,33 @@ func (c *Client) ForwardApmData(ctx context.Context) error {
 	if c.IsUnhealthy() {
 		return nil
 	}
+
+	// Wait for metadata to be available, metadata will be available as soon as
+	// the first agent data is processed.
+	select {
+	case <-ctx.Done():
+		c.logger.Debug("Invocation context cancelled, not processing any more agent data")
+		return nil
+	case data := <-c.AgentDataChannel:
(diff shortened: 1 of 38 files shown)


apmproxy/batch.go
```diff
@@ -0,0 +1,70 @@
...
+// A batch is marked as ready for flush once it reaches
+// 90% of its max size.
+func (b *BatchData) ShouldFlush() bool {
+	return len(b.agentData) >= int(float32(b.maxSize)*0.9)
+}
+
+// Reset resets the batch to prepare for new set of data
+func (b *BatchData) Reset() {
+	b.agentData = nil
+	b.agentData = make([]APMData, 0, b.maxSize)
```
axw, line 69: Can we reuse the slice? It appears that the slice won't be used after it's reset.

axw: Also, should we have a method for accessing the elements, say `func (*BatchData) Data() []APMData`? It might help clarify the expectations. e.g. we could add a comment to `Reset that says it will invalidate any values previously returned by `Data()`, similar to the comments on https://pkg.go.dev/bytes#Buffer.Bytes

apmproxy/apmserver.go
```diff
@@ -313,3 +314,64 @@ func (c *Client) WaitForFlush() <-chan struct{} {
...
+			metadata, err := ProcessMetadata(apmData)
+			if err != nil {
+				return fmt.Errorf("failed to extract metadata from agent payload %w", err)
+			}
+			c.setMetadata(metadata)
+		}
+		return c.PostToApmServer(ctx, apmData)
+	case Lambda:
+		if err := batch.Add(apmData); err != nil {
+			c.logger.Warnf("Dropping data due to error: %v", err)
```
axw, line 331: This worries me. If we are going to send a transaction for OOMing/timed-out Lambda invocations, then silently dropping data like this may skew results. Can we find a way to push back on the Lambda runtime, so it backs off and retries sending the logs later?

apmproxy/batch.go
```diff
@@ -0,0 +1,70 @@
...
+// ErrBatchFull signfies that the batch has reached full
+// capacity and cannot accept more entires.
+var ErrBatchFull = errors.New("batch is full")
+
+// BatchData represents a batch of data without metadata
+// that will be sent to APMServer. BatchData is not safe
+// concurrent access.
+type BatchData struct {
+	agentData []APMData
+	maxSize   int
```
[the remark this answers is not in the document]
lahsivjar: I have been thinking about how I am handling `agentData` here while working on buffering events. I decided to use `bytes.Buffer` instead, [here is the implementation](https://github.com/elastic/apm-aws-lambda/blob/ded2f0929533ab972226b7ffb77b534a331110a2/apmproxy/batch.go#L47). Let me know your thoughts on this, I can incorporate the changes in this PR if it sounds good.

apmproxy/batch.go
lahsivjar: > Can we reuse the slice? It appears that the slice won't be used after it's reset.

I am thinking if encoding APMData directly would be better. Please take a look at [this comment](https://github.com/elastic/apm-aws-lambda/pull/314#discussion_r982260224).

apmproxy/batch.go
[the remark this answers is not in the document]
axw: I think that looks sensible.

apmproxy/batch.go
lahsivjar: > Also, should we have a method for accessing the elements, say func (*BatchData) Data() []APMData?

I have updated the code to use `bytes.Buffer` in the batch. A method `ToAPMData` is added to get the accumulated data but now there is no way to access the elements. Let me know if the comments still needs to be addressed in the current logic.

apmproxy/apmserver.go
lahsivjar: As per the current logic, we try to send data as soon as the batch reaches 90% usage or ages to 10 seconds (from the time the first entry was added) so the case when we drop data will occur only if APM-Server has failed for more than one tries. In addition to this, pushing back on lambda in such cases sounds reasonable. I think it should be possible to do this, is it okay to address this in a separate PR?

main_test.go
```diff
@@ -692,7 +694,7 @@ func TestInfoRequestHangs(t *testing.T) {
 	lambdaServerInternals := newMockLambdaServer(t, logsapiAddr, eventsChannel, l)
 
 	eventsChain := []MockEvent{
-		{Type: InvokeStandardInfo, APMServerBehavior: Hangs, ExecutionDuration: 1, Timeout: 500},
+		{Type: InvokeStandardInfo, APMServerBehavior: Hangs, ExecutionDuration: 1, Timeout: 5},
```
lahsivjar, line 697: [**For Reviewers**] I am not exactly sure about the purpose of this test. Based on the comment, it seems to test that extension times out in case APMServer hangs. Before this PR, the event process loop was terminated by sending of `RuntimeDone` event since the configured timeout in the test was `500` however, after the implementation of pushback in case no metadata is available `RuntimeDone` event is no longer processed. I have updated the `Timeout` to 5 so that it times out based on the deadline condition.

(for ref: [the logic to terminate the event processing loop](https://github.com/elastic/apm-aws-lambda/blob/995d42614a68d43bc531e15a36bcbee27606a9c4/app/run.go#L199-L213))

apmproxy/apmserver.go
```diff
@@ -40,27 +40,20 @@ type jsonError struct {
...
-		case agentData := <-c.DataChannel:
-			if metadataContainer.Metadata == nil {
-				metadata, err := ProcessMetadata(agentData)
-				if err != nil {
-					return fmt.Errorf("failed to extract metadata from agent payload %w", err)
-				}
-				metadataContainer.Metadata = metadata
-			}
-			if err := c.PostToApmServer(ctx, agentData); err != nil {
+			return c.sendBatch(ctx)
```
axw, line 54: If the context is cancelled, then shouldn't this mean `sendBatch` should not do anything?

Looking at the sendBatch method, it calls PostToApmServer which makes an HTTP request to APM Server. The HTTP request does not use the provided context, hence why `sendBatch` doesn't get cancelled immediately.

Perhaps on context cancellation here we should use a new background context with a shutdown timeout. WDYT?

axw (commented):
Generally looks good, but I'd prefer if we could avoid having the logsapi code be aware of metadata. I think ideally logsapi would just send log events to a channel, and if that channel is full then the log handler can eventually time out and cause the Lambda platform to retry later.

apmproxy/apmserver.go
axw: Sounds reasonable - thanks.

apmproxy/batch.go
axw: It looks good now, thanks.

logsapi/route_handlers.go
```diff
@@ -25,8 +25,22 @@ import (
...
-func handleLogEventsRequest(logger *zap.SugaredLogger, logsChannel chan LogEvent) func(w http.ResponseWriter, r *http.Request) {
+func handleLogEventsRequest(
+	logger *zap.SugaredLogger,
+	logsChannel chan LogEvent,
+	metadataAvailable <-chan struct{},
+) func(w http.ResponseWriter, r *http.Request) {
 	return func(w http.ResponseWriter, r *http.Request) {
+		select {
+		case <-metadataAvailable:
+		case <-time.After(100 * time.Millisecond):
```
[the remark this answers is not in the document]
lahsivjar: Agree with the smell. I thought about this approach but went with pushback at logsapi level. The reason why I chose this approach is because of 2 reasons:
1. In my tests I was losing a bunch of data in case metadata was not available and the buffer just got lost because the function was shutdown, however, my tests were not replicating real-world usage since I sent a few requests and then stopped. Assuming that a lambda function will be invoked with a sustained throughput the chances of this case being a concern will be less. WDYT?
2. The logsapi sends logs in a batch and the channel accepts them one by one. Due to this, it is possible that a single payload of requests has some events which are ingested and some which are not. This will cause some events to be ingested multiple times. I am not sure how our stack handles duplicate events with the same timestamp so would love to discuss more on this.

apmproxy/batch.go
```diff
@@ -0,0 +1,117 @@
...
+	buf             bytes.Buffer
+	count           int
+	maxSize         int
+	maxAgeInSeconds int
+	age             int64
+}
+
+// NewBatch creates a new BatchData which can accept a
+// maximum number of entires as specified by the argument
+func NewBatch(maxSize int, maxAgeInSeconds int, metadata []byte) *BatchData {
```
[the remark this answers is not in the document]
lahsivjar: Makes sense, the thought just escaped me 😅

apmproxy/apmserver.go
lahsivjar: Ah, thanks for catching this. It is the remnant from when I was creating local batch and to avoid data-loss, flushing them on ctx done. I will fix this to not do anything on context canceled since now the batch is global.

apmproxy/batch.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	if err := b.buf.WriteByte('\n'); err != nil {
+		return err
+	}
+	if _, err := b.buf.Write(d.Data); err != nil {
+		return err
+	}
+	if b.count == 0 {
+		// For first entry, set the age of the batch
+		b.age = time.Now().Unix()
```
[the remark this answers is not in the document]
lahsivjar: TIL about monotonic clocks

logsapi/route_handlers.go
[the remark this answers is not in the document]
axw: > In my tests I was losing a bunch of data in case metadata was not available and the buffer just got lost because the function was shutdown, however, my tests were not replicating real-world usage since I sent a few requests and then stopped. Assuming that a lambda function will be invoked with a sustained throughput the chances of this case being a concern will be less. WDYT?

Sorry, I don't understand. Why does changing from explicitly blocking to blocking because the channel is full make a difference? In case I was unclear in my description before, here's roughly what I meant:

1. Receive log events here, send them on the channel; nothing is receiving from the channel initially.
2. First request comes in from the agent with metadata. This event causes the extension to start receiving logs from the channel.

> The logsapi sends logs in a batch and the channel accepts them one by one. Due to this, it is possible that a single payload of requests has some events which are ingested and some which are not. This will cause some events to be ingested multiple times. I am not sure how our stack handles duplicate events with the same timestamp so would love to discuss more on this.

Excellent point. We don't do any deduplication. In theory we could set a document ID for logs and silently discard duplicates, but APM Server does not currently support specifying document IDs like this.

I suppose we would need to handle batches atomically to do what I described. Or implement deduplication, but that's not such an easy lift.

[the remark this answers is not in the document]
lahsivjar: > Sorry, I don't understand. Why does changing from explicitly blocking to blocking because the channel is full make a difference?

Apologies, I didn't explain well. The issue happens when the execution environment shuts down before it receives the metadata. In case of explicit blocking, there would be nothing in the channel unless metadata is available however, in the other case the channel can have some data and when the execution env shuts down without getting metadata the data in the channel will be lost. In the explicit blocking case the metadata can be processed by another execution env.

> I suppose we would need to handle batches atomically to do what I described.

One approach can be to create a queue that will allow us to reserve a number of slots. I can give it a try. Should we do it in a separate PR? In this PR, if the assumption in the first point sounds good, we can remove the metadata handling in logsapi and fix the atomic handling of the batch in another PR.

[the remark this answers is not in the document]
axw: > however, in the other case the channel can have some data and when the execution env shuts down without getting metadata the data in the channel will be lost

OK, I think I understand now. Because we're using a buffered channel, logs may end up in the channel but could hypothetically never be consumed. If we don't accept the logs into the channel, then the logs might later be re-delivered to another instantiation of the extension.

> In my tests I was losing a bunch of data in case metadata was not available and the buffer just got lost because the function was shutdown, however, my tests were not replicating real-world usage since I sent a few requests and then stopped. Assuming that a lambda function will be invoked with a sustained throughput the chances of this case being a concern will be less. WDYT?

I don't think we should ever assume that there will be sustained throughput. However...

If we never receive metadata before shutdown, wouldn't that imply an instrumentation problem? And in this case we would never do anything with the logs anyway, even if they are sent to the next instantiation of the extension?

apmproxy/apmserver.go
lahsivjar: Fixed this to not do anything on context done. The extension will rely on `FlushAPMData` to take care of anything in the batch.

logsapi/route_handlers.go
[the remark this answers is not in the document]
lahsivjar: > If we never receive metadata before shutdown, wouldn't that imply an instrumentation problem?

I am not sure about this. IIUC, agent will send data via HTTP call and if the lambda function is small it might lead to the call not being successful and the function returning. I observed this behaviour while testing.

<details>
  <summary>Lambda function used for testing</summary>
  
```
import json
from elasticapm import capture_serverless

@capture_serverless()
def lambda_handler(event, context):
    print("This is function log", context.aws_request_id)
    return {
        'statusCode': 200,
        'body': json.dumps('Hello from Lambda!'+context.aws_request_id)
    }
```
</details>

apmproxy/batch.go
```diff
@@ -0,0 +1,118 @@
...
+	metadataBytes int
+	buf           bytes.Buffer
+	count         int
+	age           time.Time
+	maxSize       int
+	maxAge        time.Duration
+}
+
+// NewBatch creates a new BatchData which can accept a
+// maximum number of entires as specified by the argument
```
simitt, line 55:
```suggestion
// maximum number of entries as specified by the argument
```

apmproxy/client.go
```diff
@@ -46,13 +46,15 @@ const (
 	defaultDataForwarderTimeout time.Duration = 3 * time.Second
 	defaultReceiverAddr                       = ":8200"
 	defaultAgentBufferSize      int           = 100
+	defaultMaxBatchSize         int           = 50
+	defaultMaxBatchAge          time.Duration = 10 * time.Second
```
simitt, line 50: This seems like a long period of time for the extension. How did you come up with 10 seconds?

simitt (commented):
Only left mostly cosmetic comments for now.

apmproxy/client.go
lahsivjar: Nice catch. Considering the nature of the lambda function the duration is definitely too big. My first intention of adding `age` factor to the batch was to account for functions with very low throughput. I got caught up in keeping the batch to be of good size.

Ideally, I think this value should be chosen based on the type of workload. So maybe we can expose configuration options for this(?)

For a good default, we need to balance the batch size and the age. Lambda provides a max time of 15 minutes for a function to run but I think most functions would be < 1 second(?). So, we can keep the default as 2 seconds in aim to flush everything by the second invocation maximum. What is your recommendation here?

apmproxy/batch.go
lahsivjar: I keep misspelling entries 😅

apmproxy/client.go
simitt: 2 seconds sounds reasonable. We might need to expose a config option for this at some point. Right now, given that the logs collection is in Tech Preview, I'd rather suggest to disable the collection if the `maxBatchAge` becomes an issue for customers.

apmproxy/client.go
```diff
@@ -46,13 +46,15 @@ const (
 	defaultDataForwarderTimeout time.Duration = 3 * time.Second
 	defaultReceiverAddr                       = ":8200"
 	defaultAgentBufferSize      int           = 100
+	defaultMaxBatchSize         int           = 50
```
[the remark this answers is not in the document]
lahsivjar: @simitt I am not sure if it would work in every case since I can't think of a direct correlation between buffer size and batch size. Due to this, I believe, giving only buffer size will make the configuration a bit more complicated.

felixbarny:
I've not been following this too closely so sorry if that was mentioned already. The logs API itself is already buffered. See also https://docs.aws.amazon.com/lambda/latest/dg/runtimes-logs-api.html#runtimes-logs-api-buffering. Could we just rely on this config rather than the user having to worry about buffering configs in both the lambda logs API settings and the APM extension settings?

lahsivjar:
@felixbarny This PR aims to utilize lambda APIs buffering instead of buffering of events in the extension. So, for lambda logs, if the extension is not able to consume the events then we return a 5xx to the lambda API making lambda buffer the logs (or drop them if buffer settings are exceeded). The config `ELASTIC_APM_LAMBDA_AGENT_DATA_BUFFER_SIZE` is for buffering events from APM-agents. The batch config introduced in this PR is for the number of events in the payload to APM-Server.

Let me know if this doesn't address your concern properly.

felixbarny:
Thanks for bringing me up to speed. Makes a lot of sense!

lahsivjar:
@axw I have updated the PR to avoid `logsapi` being aware of the metadata availability. Pushback to lambda in case of errors is not complete yet, I will create a separate issue to track that. The below diagram gives a gist of implementation in the current PR:
 
pr314

To summarize:

**What current PR does?**
1. Batches events for the request to APM-Server
2. Buffers event in the extension until metadata is available
3. If the buffer is full, the extension pushes back on lambda API

**What current PR doesn't do?**
1. Push back on Lamba logs API for ALL cases (including metadata not available)

apmproxy/apmserver.go
```diff
@@ -40,28 +40,40 @@ type jsonError struct {
...
+				return err
 			}
-			if err := c.PostToApmServer(ctx, agentData); err != nil {
-				return fmt.Errorf("error sending to APM server, skipping: %v", err)
+		case data := <-c.LambdaDataChannel:
+			if err := c.forwardLambdaData(ctx, data); err != nil {
+				return err
 			}
 		}
 	}
```
axw, line 79: Do we still need the `metadataAvailable` channel? And if we do, does it need to be threaded through the entire application? IIANM we only care about it in the apmproxy package now.

IIANM, I think we could get rid of the channel and simplify this code at the same time by using a pattern like this:

```go
if c.IsUnhealthy() {
    return nil
}

// Process lambda and agent data after metadata is available
var lambdaDataChannel chan APMData
for {
    select {
    case <-ctx.Done():
        c.logger.Debug("Invocation context cancelled, not processing any more agent data")
        return nil
    case data := <-c.AgentDataChannel:
        if err := c.forwardAgentData(ctx, data); err != nil {
            return err
        }
        // Metadata is now available, so we can start processing events from the Lambda platform.
        lambdaDataChannel = c.LambdaDataChannel
    case data := <-c.LambdaDataChannel:
        if err := c.forwardLambdaData(ctx, data); err != nil {
            return err
        }
    }
}
```

apmproxy/apmserver.go
```diff
@@ -74,15 +86,48 @@ func (c *Client) FlushAPMData(ctx context.Context) {
...
+			case data := <-c.AgentDataChannel:
+				if err := c.forwardAgentData(ctx, data); err != nil {
+					c.logger.Errorf("Error sending to APM Server, skipping: %v", err)
+				}
+			default:
+				c.logger.Debug("Flush ended for agent data - no data in buffer")
+				return
+			}
+		}
+	}()
```
axw, line 105:
```suggestion
	for i := len(c.AgentDataChannel); i > 0; i-- {
		data := <-c.AgentDataChannel
		if err := c.forwardAgentData(ctx, data); err != nil {
			c.logger.Errorf("Error sending to APM Server, skipping: %v", err)
		}
	}
```

apmproxy/apmserver.go
```diff
@@ -74,15 +86,48 @@ func (c *Client) FlushAPMData(ctx context.Context) {
...
+
+	// If metadata still not available then fail fast
+	select {
+	case <-c.metadataAvailable:
+	default:
+		c.logger.Warnf("Metadata not available at flush, skipping sending lambda data to APM Server")
+		return
+	}
```
axw, line 113:
```suggestion
	// If metadata still not available then fail fast
	if c.batch == nil {
		c.logger.Warnf("Metadata not available at flush, skipping sending lambda data to APM Server")
		return
	}
```

apmproxy/batch_test.go
```diff
@@ -0,0 +1,78 @@
...
+	assert.True(t, b.ShouldShip())
+}
+
+func TestShouldShip_ReasonAge(t *testing.T) {
+	b := NewBatch(10, time.Second, []byte(metadata))
+
+	assert.False(t, b.ShouldShip())
+	require.NoError(t, b.Add(APMData{}))
+
+	<-time.After(2 * time.Second)
```
axw, line 73:
```suggestion
	time.Sleep(2 * time.Second)
```

Could we reduce the duration to 1s+1ms, like in TestForwardApmData? I think it should be safe now that we're using `time.Since`, hence monotonic clock.

apmproxy/batch_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+	// Should be ready for send after 3 seconds
```
axw, line 75: comment doesn't reflect the duration above - maybe just say "ready to send now"

apmproxy/apmserver_test.go
```diff
@@ -586,6 +604,134 @@ func TestContinuedAPMServerFailure(t *testing.T) {
...
+	}
+
+	// Send lambda logs API data
+	var expected bytes.Buffer
+	expected.WriteString(metadata)
+	// Send multiple lambda logs to batch data
+	for i := 0; i < 5; i++ {
+		if i == 4 {
+			// Wait for batch age to make sure the batch is mature to be sent
+			<-time.After(maxBatchAge + time.Millisecond)
```
axw, line 674:
```suggestion
			time.Sleep(maxBatchAge + time.Millisecond)
```

axw (commented):
Looks good overall. I'd like to get rid of the `metadataAvailable` channel if possible. I _think_ it's something that can be encapsulated within the apmproxy package.

apmproxy/apmserver.go
lahsivjar: Love this solution

apmproxy/apmserver.go
```diff
@@ -74,15 +76,37 @@ func (c *Client) FlushAPMData(ctx context.Context) {
...
+
+	// Flush lambda data
 	for {
 		select {
-		case agentData := <-c.DataChannel:
-			c.logger.Debug("Flush in progress - Processing agent data")
-			if err := c.PostToApmServer(ctx, agentData); err != nil {
+		case <-ctx.Done():
+			c.logger.Debugf("Failed to flush completely, may result in data drop")
```
axw, line 98:
```suggestion
			c.logger.Debugf("Failed to flush completely, may result in data drop")
			return
```

IIANM we must return here or we'll enter an infinite loop, as we'll only hit the default case if none of the other cases can be run.

axw (approved):
Just noticed one other issue, otherwise LGTM.

apmproxy/apmserver.go
lahsivjar: OMG, thanks for catching this.

lahsivjar: Fixed.

apmproxy/client.go
lahsivjar: @simitt 
> I'd rather suggest to disable the collection if the maxBatchAge becomes an issue for customers.

FYI, the batching also applies to platform metrics. Is this as per your expectations?

PR #310, opened 2022-09-19
ci: add smoke tests integration

kruskall:
Initial work to integrate smoke testing scripts into CI. 

Mostly inspired by the work in the mutating webhook and APM server, add make tasks to make testing easier and provide a target for CI to discover, run and cleanup the environment.

The tasks are flexible and support muiltiple smoke tests (currently there's only one).

Related to https://github.com/elastic/apm-aws-lambda/issues/217

PR #329, opened 2022-10-08
ci: add smoke testing job definition

kruskall:
Companion PR to https://github.com/elastic/apm-aws-lambda/pull/310

Add job definition

### DIFF
--- a/.ci/jobs/apm-aws-lambda-smoketests-mbp.yml
+++ b/.ci/jobs/apm-aws-lambda-smoketests-mbp.yml
@@ -1,0 +1,41 @@
+---
+- job:
+    name: library/apm-aws-lambda-smoketests-mbp
+    display-name: AWS Lambda extension Smoke Tests
+    description: AWS Lambda extension Smoke Tests
+    project-type: multibranch
+    concurrent: true
+    script-path: .ci/smoke-tests.groovy
+    scm:
+    - github:
+        branch-discovery: no-pr
+        discover-pr-forks-strategy: merge-current
+        discover-pr-forks-trust: permission
+        discover-pr-origin: merge-current
+        discover-tags: false
+        head-filter-regex: '(main|PR-.*|v\d+\.\d+\.\d+.*)'
+        notification-context: 'apm-ci'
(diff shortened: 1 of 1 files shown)


.ci/jobs/apm-aws-lambda-smoketests-mbp.yml
```diff
@@ -0,0 +1,41 @@
...
+    concurrent: true
+    script-path: .ci/smoke-tests.groovy
+    scm:
+    - github:
+        branch-discovery: no-pr
+        discover-pr-forks-strategy: merge-current
+        discover-pr-forks-trust: permission
+        discover-pr-origin: merge-current
+        discover-tags: false
+        head-filter-regex: '(main|PR-.*|v\d+\.\d+\.\d+.*)'
```
v1v, line 16: If `discover-tags: false`, then `|v\d+\.\d+\.\d+.*` won't be needed, as far as I know that's the regex for the tags

The below suggestion, will discover all the PRs and changes in `main`
```suggestion
        head-filter-regex: '(main|PR-.*)'
```

v1v (commented):
There are some questions regarding when this particular job should run, see https://github.com/elastic/apm-aws-lambda/pull/310#discussion_r975408468

Those answers might imply some changes in this PR.

At the moment, the current changes will imply:
- Run always for every push in any PR or main.

If you can clarify when this job should run we can configure it accordingly.

OTOH, I normally like to keep support for PRs, and disable the automation for triggering the build for those PRs if needed. Hence, it's possible to run the pipeline for any PR manually (using the Jenkins UI, or even a GitHub comment), that's  handy when testing changes related to the pipeline itself.

kruskall:
> OTOH, I normally like to keep support for PRs, and disable the automation for triggering the build for those PRs if needed. Hence, it's possible to run the pipeline for any PR manually (using the Jenkins UI, or even a GitHub comment), that's handy when testing changes related to the pipeline itself.

Agree, I think it's okay to only run it on main.

v1v (approved):

PR #332, opened 2022-10-18
Change default send strategy to `background`

AlexanderWert:


simitt (commented):
Please add a changelog entry.

Issue #315, opened 2022-09-23
POC: Create transaction in lambda extension and record error results


simitt:
This enhancement issue to build a POC to investigate creating transaction events on behalf of the Lambda function, and report an error result (e.g. "OutOfMemory") in `transaction.result` in case an error occurs. 

Related go agent work: https://github.com/elastic/apm-agent-go/issues/1323

** Needs more definition **

axw:
My expectation is that eventually, the transaction document will capture:
 - the error result, e.g. `OutOfMemory` or `TimedOut`
 - Lambda/`faas` attributes and metrics

The document would look something like

```json
{
  "transaction": {
    "name": "...",
    "result": "OutOfMemory"
  },
  "faas": {
    "execution": "...",
    "id": "...",
    "coldstart": true,
    "duration": 123.456,
    "billed_duration": 200,
    "coldstart_duration": 100,
    "timeout": 5000,
  },
  "system.memory.total": 536870912,
  "system.memory.actual.free": 0,
}
```

Later, the `faas.*` and `system.*` fields will be aggregated in transaction metrics. This should ideally be generalised to request-scoped metrics.

PR #336, opened 2022-10-25
Remove usage of APMData within logsapi

lahsivjar:
Currently, the package `apmproxy` is required by the package `logsapi` due to the abuse of `APMData` struct. The PR refactors the struct so that it's usage is contained within `apmproxy`.

Part of #315


axw (approved):
Nice, thank you.

PR #337, opened 2022-10-26
Create proxy transasction if not reported by agent

lahsivjar:
## Motivation

Create proxy transactions if agent is not able to report the transaction due to unexpected errors like `OutOfMemory`.

## Related issues

Part of #315 

## Testing

1. Create a lambda function that causes an OOM and run it a few times.
2. Observe on elastic to assert that a transaction for the function invocation is reported.


logsapi/option.go
```diff
@@ -55,3 +58,11 @@ func WithSubscriptionTypes(types ...SubscriptionType) ClientOption {
 		c.logsAPISubscriptionTypes = types
 	}
 }
+
+// WithBatch configures a batch to be used for batching data
+// before sending to APM Server.
+func WithBatch(batch *accumulator.Batch) ClientOption {
```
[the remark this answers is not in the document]
lahsivjar: > If we need to pass some kind of callback in, maybe make it an interface (or interfaces)? That would remove the direct dependency on accumulator.

Thanks for the tip, I have added an interface to remove accumulator dependency on logsapi.

> I wonder if we should just move ProcessLogs out of this package though, and into package app.

This sounds worth trying. I can give it a try with the PR to remove the buffer from logsapi or do you think we should address it in this PR itself?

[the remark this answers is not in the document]
axw: Doing it in a separate PR sounds good.

apmproxy/receiver.go
```diff
@@ -164,3 +159,20 @@ func (c *Client) handleIntakeV2Events() func(w http.ResponseWriter, r *http.Requ
 		}
 	}
 }
+
+// URL: http://server/register/transaction
+func (c *Client) handleTransactionRegistration() func(w http.ResponseWriter, r *http.Request) {
+	return func(w http.ResponseWriter, r *http.Request) {
+		txnID := r.Header.Get("x-elastic-transaction-id")
+		traceID := r.Header.Get("x-elastic-trace-id")
```
axw, line 167: I think we should be receiving the full `traceparent` and optionally `tracestate` here. `traceparent` is needed for setting the parent ID, in case the Lambda invocation is not the trace root. `tracestate` is needed for setting the `transaction.sample_rate` field.

accumulator/invocation.go
```diff
@@ -0,0 +1,106 @@
...
+func (inc *Invocation) Finalize(status string) {
+	if inc.TransactionID == "" {
+		return
+	}
+	if rootTxnIdx := inc.findRootTxn(); rootTxnIdx == -1 {
+		inc.createProxyTxn(status)
+	}
+}
+
+func (inc *Invocation) createProxyTxn(status string) int {
```
axw, line 64: Maybe use `go.elastic.co/apm/v2/model.Transaction` here?

accumulator/invocation.go
```diff
@@ -0,0 +1,106 @@
...
+	var w fastjson.Writer
+	w.RawString(`{"transaction":{"id":`)
+	w.String(inc.TransactionID)
+	w.RawString(`,"trace_id":`)
+	w.String(inc.TraceID)
+	w.RawString(`,"result":`)
```
axw, line 70: Should we also set `outcome`?

accumulator/invocation.go
```diff
@@ -0,0 +1,106 @@
...
+	if rootTxnIdx := inc.findRootTxn(); rootTxnIdx == -1 {
+		inc.createProxyTxn(status)
+	}
+}
+
+func (inc *Invocation) createProxyTxn(status string) int {
+	var w fastjson.Writer
+	w.RawString(`{"transaction":{"id":`)
+	w.String(inc.TransactionID)
+	w.RawString(`,"trace_id":`)
```
axw, line 68: We might also need to set `parent_id`

accumulator/invocation.go
```diff
@@ -0,0 +1,106 @@
...
+	w.String(inc.TraceID)
+	w.RawString(`,"result":`)
+	w.String(status)
+	w.RawString("}}")
```
axw, line 72: Other fields I think we should be setting:
 - `name` (to what I'm not sure... I think ideally we would extract the payload and use it to set the name)
 - `type` ("request"? ideally we would set it according to the trigger type as described in https://github.com/elastic/apm/blob/main/specs/agents/tracing-instrumentation-aws-lambda.md)
 - `timestamp` (to the invocation timestamp?)

accumulator/batch.go
```diff
@@ -0,0 +1,249 @@
...
+	// Set metadata if not already set
+	if b.metadataBytes == 0 {
+		b.metadataBytes, _ = b.buf.Write(data[0])
+	}
+	// Should we identify the target transaction and cache only that?
+	for i := 1; i < len(data); i++ {
+		if d := data[i]; len(d) > 0 {
+			inc.agentData = append(inc.agentData, d)
+		}
+	}
```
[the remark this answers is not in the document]
lahsivjar: > That seems sensible to me, since we're not enriching the events in the extension. 

Does the APM-Server and ES have any constraints around the order of events. I was thinking that if a transaction event is not reported for an invocation but other events are reported then will shipping the other events before the txn cause an unexpected state? (In this case a proxy transaction will be created and will be shipped at a later time but other events for the same invocation might end up earlier)

> That would imply that we would revert to sending the events through to APM Server immediately, right?

Since we will need to read the body of the payload anyway, I was thinking to have a single batch that will be shipped to APM-Server instead of shipping agent data individually. The batch will have a metadata and a combination of lambda events + agents events. The batch will be shipped to APM Server according to the batch maturity settings in the extension (age, size, shutdown event). In this case, for agent payload, the extension will:

1. Read the payload from agents line by line and strip metadata
2. Read the type of event on each line; if the event is the target transaction we will mark the invocation as finalized otherwise we will wait for the `runtimeDone` event from lambda logs (depending on the constraints of the order of events this can also be optimized).

WDYT?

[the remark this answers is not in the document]
axw: > Does the APM-Server and ES have any constraints around the order of events.

Transactions do not have to precede the other events. In general they won't; enclosed events (such as spans) will typically be ended before their enclosing transaction, and thus be sent to APM Server first.

> Since we will need to read the body of the payload anyway, I was thinking to have a single batch that will be shipped to APM-Server instead of shipping agent data individually. The batch will have a metadata and a combination of lambda events + agents events.

OK, I think that sounds good.

accumulator/invocation.go
lahsivjar: What should be the value for the `outcome`? I was a bit confused between what these fields indicates.

Initially, I chose to set only `outcome`, as the description for `transaction.result` felt not completely aligned with the [description of the result field](https://www.elastic.co/guide/en/apm/guide/8.4/exported-fields-apm-transaction.html), however, I can only find `event.outcome` in the documentations and no `transaction.outcome` but we do seem to [export transaction.outcome from apm-agent-go](https://github.com/elastic/apm-agent-go/blob/4d07ba809e943acb94c347517cfbc1d760a65297/model/model.go#L276-L277).

apmproxy/receiver.go
lahsivjar: I started with just sending a JSON payload with all the fields available during its creation however, it felt a bit too much to encode JSON in the agent and decode it again here. Does it make sense to get `parentID` and `sample_rate` as headers rather than in the request body or do you foresee that in the future we might need more data from the agent?

accumulator/invocation.go
axw: It would be set to the string "failure" for failed transactions. Otherwise I think we would omit it.

> Initially, I chose to set only outcome, as the [description for transaction.result](https://www.elastic.co/guide/en/apm/guide/8.4/exported-fields-apm-transaction.html) felt not completely aligned with what we were setting, however, I can only find event.outcome in the documentations and no transaction.outcome but we do seem to [export transaction.outcome from apm-agent-go](https://github.com/elastic/apm-agent-go/blob/4d07ba809e943acb94c347517cfbc1d760a65297/model/model.go#L276-L277).

`transaction.outcome` in intake-v2 results in the `event.outcome` field being set in docs. This field is used for calculating the rather course-grained "failed transaction rate" metric.

`transaction.result` is free-form, and can be used to group transactions by a more specific error reason. This is where we would set "timeout", "oom", etc.

apmproxy/receiver.go
axw: If we were going to future-proof it, I think we would be better off just passing through the whole Lambda payload and letting the extension pull the headers out. That way the extension can also pull other information out for use in the transaction it creates (e.g. for `transaction.type` and `transaction.name`); and perhaps also to enrich transactions coming from agents.

The agent can still generate `transaction.id` and pass that through to the extension in a header -- though I do wonder if we could derive a `transaction.id` from the AWS request ID. The Lambda payload would just be passed through as-is, no additional encoding by the agent.

accumulator/batch.go
```diff
@@ -0,0 +1,249 @@
...
+	age                         time.Time
+	maxSize                     int
+	maxAge                      time.Duration
+	currentlyExecutingRequestID string
+
+	// TODO: @lahsivjar remove requirements of a mutex; currently it is
+	// required because the invocations need to be accessed from logsapi
+	// as the processed log output of logsapi doesn't have the necessary
+	// information.
+	mu sync.RWMutex
```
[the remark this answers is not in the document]
lahsivjar: All fields currently require a mutex since the agent data is added to the same batch. I have moved the mutex accordingly.

accumulator/batch.go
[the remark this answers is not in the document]
lahsivjar: I have updated the code to not cache any agent data. The logic checks if it has seen the root txn or not; in both cases it appends the event to the batch. If the txn is not seen until shutdown or `runtimeDone` for that invocation, we create a proxy txn.

apmproxy/receiver.go
lahsivjar: I have updated the code to accept the fully encoded JSON for the partial transaction and in the current logic I append `transaction.{result,outcome}` to the JSON so all fields set by the agent are respected when creating a proxy txn.

accumulator/invocation.go
lahsivjar: Now, I have updated the code so that agent will send the fully encoded partial transaction. So, we can just set the required fields without decoding the JSON. Let me know how this sounds to you.

accumulator/invocation.go
lahsivjar: Updated the logic to use the fully encoded JSON for the partially created txn in the agent. All these fields (at least for the GoAgent) will be set.

accumulator/batch.go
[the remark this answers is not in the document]
axw: Looks good!

apmproxy/receiver.go
axw: That looks good. If we're passing a partial transaction rather than the Lambda invocation payload, do we still need to send the transaction ID as a separate header? Can we pull the transaction ID out of the transaction when the register endpoint is called?

What do you think about specifying a "vnd" Content-Type (e.g. `application/vnd.elastic.apm.transaction+json`), to allow for changes later in case we do want to send and parse the Lambda payload in the extension?

accumulator/invocation.go
axw: Looks good, thanks.

axw (commented):
This looks good to me now.

The approach you've taken of sending a partial transaction is quite similar to what @felixbarny suggested with the `transaction-start` event, and looks good. I still think it could be useful to send the whole payload to the extension and have it parse it, but that's a lot more work and can be deferred until/unless we need it. Maybe specifying a Content-Type would help us handle that change gracefully.

apmproxy/receiver.go
lahsivjar: > If we're passing a partial transaction rather than the Lambda invocation payload, do we still need to send the transaction ID as a separate header? Can we pull the transaction ID out of the transaction when the register endpoint is called?

Yes, we can. I did it in the draft version before moving to header approach and somehow missed it here. I have fixed this.

> What do you think about specifying a "vnd" Content-Type (e.g. application/vnd.elastic.apm.transaction+json), to allow for changes later in case we do want to send and parse the Lambda payload in the extension?

This sounds like a good idea to me. I have added a check in the extension for `content-type` header.

accumulator/batch.go
```diff
@@ -0,0 +1,275 @@
...
+// AddAgentData adds a data received from agent. For a specific invocation
+// agent data is always received in the same invocation.
+func (b *Batch) AddAgentData(apmData APMData) error {
+	raw, err := GetUncompressedBytes(apmData.Data, apmData.ContentEncoding)
+	if err != nil {
+		return err
+	}
+	// A request body can either be empty or have a ndjson content with
+	// first line being metadata.
+	data := bytes.Split(raw, []byte("\n"))
```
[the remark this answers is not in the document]
axw: `bytes.Split` creates a slice, causing allocations. It might not be a big deal, but in my experience using a scanner or using `bytes.Index` or `bytes.Cut` has less overhead.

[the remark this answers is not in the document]
lahsivjar: Thanks for the tip, initially my thought process was to use Split since the returned slice used in Split would be a slice of subslices and would end up on stack but I see that the Split would also do 2 scans over the input. I have updated the code to use `bytes.Cut`.

accumulator/batch.go
```diff
@@ -0,0 +1,283 @@
...
+	}
+	if b.count == 0 {
+		// For first entry, set the age of the batch
+		b.age = time.Now()
+	}
+	b.count++
+	return nil
+}
+
+func findEventType(body []byte) []byte {
```
axw, line 268: Should we make this return a string? We're converting the result to a string in all cases where it's called.

Or alternatively, change it to `func isTransactionEvent(body []byte) bool`, to avoid allocating due to conversion from []byte to string.

axw (approved):
LGTM, just one other minor comment.

accumulator/batch.go
lahsivjar: SGTM, refactored the code with `isTransactionEvent`

carsonip:
Tested extension on main branch (commit 3032883ee8622076fc3bbc557d44ad5542d48dd2), with ESS 8.6.0 and a PoC that takes a param to decide whether it times out. Not all timed out calls are recorded in ES. Will have to investigate.

carsonip:
Tested extension on main branch with ESS 8.6.0 and PoC with go agent on branch https://github.com/lahsivjar/apm-agent-go/tree/1323-lambdatxn

This PR is working well. Timed out requests will report `event.outcome: failure`, with the exception that if the first invocation is a failure in an execution environment, the data will be dropped due to missing metadata. The discovery is shared with @lahsivjar @felixbarny @axw

PR #343, opened 2022-11-14
fix: do not return status 200 on server info request when unreachable

kruskall:
When proxying info requests to apm server we're loggin an error if the server is unreachable.
The ErrorHandler should also write a status code back to avoid returning a Status 200.
Add ModifyResponse to log a debug message with the body on unexpected responses and let the ErrorHandler proxy the status code back to the agent.
A StatusBadGateway is returned for generic errors.

### DIFF
--- a/apmproxy/receiver.go
+++ b/apmproxy/receiver.go
@@ -73,6 +73,12 @@ func (c *Client) Shutdown() error {
 	return c.receiver.Shutdown(ctx)
 }
 
+type unexpectedStatusError int
+
+func (e unexpectedStatusError) Error() string {
+	return fmt.Sprintf("unexpected status code: %d", int(e))
+}
+
 // URL: http://server/
 func (c *Client) handleInfoRequest() (func(w http.ResponseWriter, r *http.Request), error) {
 	// Init reverse proxy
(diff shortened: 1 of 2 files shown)


apmproxy/receiver.go
```diff
@@ -87,9 +93,39 @@ func (c *Client) handleInfoRequest() (func(w http.ResponseWriter, r *http.Reques
 	customTransport.ResponseHeaderTimeout = c.client.Timeout
 	reverseProxy.Transport = customTransport
 
+	reverseProxy.ModifyResponse = func(rsp *http.Response) error {
+		if rsp.StatusCode != http.StatusOK {
```
axw, line 97:
```suggestion
		if rsp.StatusCode == http.StatusOK {
			return nil
		}
```

Nit: return early to improve readability of the following code.

apmproxy/receiver.go
```diff
@@ -87,9 +93,39 @@ func (c *Client) handleInfoRequest() (func(w http.ResponseWriter, r *http.Reques
...
+			b, err := io.ReadAll(rsp.Body)
+			if err != nil {
+				c.logger.Warnf("failed to read version response body: %v", err)
+				b = []byte{}
```
axw, line 101: Is there any harm in using the result of `io.ReadAll` below? It's likely to be nil, which is fine to type-assert to a string. If it's not nil, it may be interesting to log that.

apmproxy/receiver.go
```diff
@@ -87,9 +93,39 @@ func (c *Client) handleInfoRequest() (func(w http.ResponseWriter, r *http.Reques
...
+			}
+
+			c.logger.Debugf("failed to query version from the APM server: response body: %s", string(b))
+
+			// Return an error to the ErrorHandler that will proxy the status code back to the agent.
+			return fmt.Errorf("failed to query version from the APM server: response status: %w", unexpectedStatusError(rsp.StatusCode))
```
axw, line 107: Why is this necessary? If we return nil here, then won't rsp.StatusCode already be set to whatever APM Server returned?

apmproxy/receiver.go
kruskall: The `ModifyResponse` func has been removed so this doesn't apply anymore.

apmproxy/receiver.go
kruskall: The `ModifyResponse` func has been removed so this doesn't apply anymore.

apmproxy/receiver.go
kruskall: Good point! This was added to increase visibility in the lambda extension, mostly for agents that do not log error reponse code/body.

Looking back at this, it's probably not appropriate for the lambda extension to touch the response. It should proxy it as is and leave that responsibility to the agent.

I've update the code to remove the the `ModifyResponse` func :+1:

kruskall:
Sorry for the late reply. This has been updated!

apmproxy/receiver.go
```diff
@@ -91,6 +97,19 @@ func (c *Client) handleInfoRequest() (func(w http.ResponseWriter, r *http.Reques
...
 		c.logger.Errorf("Error querying version from the APM server: %v", err)
+
+		uErr := new(unexpectedStatusError)
+		if errors.As(err, uErr) {
+			status := int(*uErr)
+
+			// Proxy the status code to the agent.
+			w.WriteHeader(status)
+			return
+		}
```
axw, line 108:
```suggestion
```

AFAICS, this code path is completely dead - since ModifyResponse was returned, err will never be unexpectedStatusError.

apmproxy/receiver.go
```diff
@@ -73,6 +73,12 @@ func (c *Client) Shutdown() error {
 	return c.receiver.Shutdown(ctx)
 }
 
+type unexpectedStatusError int
+
+func (e unexpectedStatusError) Error() string {
+	return fmt.Sprintf("unexpected status code: %d", int(e))
+}
+
```
axw, line 81:
```suggestion
```

axw (approved):

PR #344, opened 2022-11-17
Handle txn registration if invocation registration is late

lahsivjar:
## What the PR does?

- Transaction registration can happen before the invocation is registered. Handle such cases by keeping the payload of transaction registration in a cache and handling it when the invocation is registered.
- Transaction registration will not have the duration of the transaction or if it does have it, the value will be a default value. We attempt to calculate the transaction duration based on the invocation timestamp, function deadline(for timeouts), OR `runtimeDone` log timestamp for agent issues.
- Increases the timeout for the deadline of extension from 100ms to 200ms.

Related to https://github.com/elastic/apm-agent-go/issues/1323

### DIFF
--- a/accumulator/batch.go
+++ b/accumulator/batch.go
@@ -60,11 +60,16 @@ type Batch struct {
 	buf bytes.Buffer
 	// invocations holds the data for a specific invocation with
 	// request ID as the key.
-	invocations                 map[string]*Invocation
-	count                       int
-	age                         time.Time
-	maxSize                     int
-	maxAge                      time.Duration
+	invocations map[string]*Invocation
+	count       int
+	age         time.Time
+	maxSize     int
+	maxAge      time.Duration
+	// currentlyExecutingRequestID represents the request ID of the currently
+	// executing lambda invocation. The ID can be set either on agent init or
+	// when extension receives the invoke event. If the agent hooks into the
+	// invoke lifecycle then it is possible to receive the agent init request
+	// before extension invoke is registered.
 	currentlyExecutingRequestID string
 }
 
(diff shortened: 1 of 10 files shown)


accumulator/batch.go
```diff
@@ -86,28 +87,43 @@ func (b *Batch) RegisterInvocation(
...
 	b.mu.Lock()
 	defer b.mu.Unlock()
 	i, ok := b.invocations[b.currentlyExecutingRequestID]
 	if !ok {
-		return fmt.Errorf("invocation for requestID %s does not exist", b.currentlyExecutingRequestID)
-	}
-	if !isTransactionEvent(payload) {
-		return errors.New("invalid payload")
+		// It is possible that the invocation is registered at a later time
+		b.invocations[waitingInvocationKey] = &Invocation{
```
axw, line 122: Should we require the agent to send the Lambda request ID along with the transaction ID and payload? Then it can always match up on request ID, and remove currentlyExecutingRequestID.

axw (commented):
Looks good, just one question.

accumulator/batch.go
lahsivjar: That sounds plausible. I was motivated by keeping the lifecycle simple by exposing less arguments to the agent but I think sending requestID (maybe as a header?) will be better. Let me make the adjustments.

lahsivjar: I have updated the PR to accept request ID as a header `x-elastic-aws-request-id`. I didn't remove `currentlyExecutingRequestID` because I think adding the constraint for all agent data to have request ID as a header might not be a good idea but adding the requirement to the txn registration API should be okay. Additionally, the `currentlyExecutingRequestID` can be set reliably in both agent init or register invocation stages to ensure a correct value for later calls. Let me know your thoughts on it.

axw: Right, I forgot that we still need to associate events with the invocation. Looks good, thanks.

axw (approved):

Issue #334, opened 2022-10-20
Extension drops platform report metrics under certain conditions


lahsivjar:
**APM AWS Lambda extensiom version** (`1.1.0` as well as current `main`):

**Describe the bug**
- [x] Extension drops the logs if the platform report metrics are reported in an invocation that does not immediately succeed the invocation generating the platform report ([relevant code](https://github.com/elastic/apm-aws-lambda/blob/ea8baf48e6b0665bc251520606ec4629a00208b2/logsapi/event.go#L90-L101)).
- [x] Extension **always** drops the last `platform.report` metric for an execution environment.
~~- [ ] #360~~ (the logs API doesn't report this at all; further discussion can continue on the dedicated issue)

**To Reproduce**
1. Run a lambda function along with the extension.
2. Invoke the function a bunch of times.
3. Observe cloudwatch logs for a warning message `report event request id didn't match the previous event id`.
4. Observe the number of metrics with `faas.billed_duration : *` to be equal to `number of invocations - number of coldstarts`.

**Expected behavior**
Extension doesn't drop platform report metrics and sends it to APM-Server.

**Provide logs (if relevant)**:

<img width="1411" alt="Screenshot 2022-10-20 at 12 40 10 PM" src="https://user-images.githubusercontent.com/1427601/196858457-e35ca0f5-9657-426e-a174-111a856ac036.png">
^ Logs from cloudwatch showing dropped platform reports (the logs is observed using a lambda extension from a dev branch)

PR #347, opened 2022-11-23
On shutdown wait for lambda logs API to report the final platform report metrics

lahsivjar:
## Motivation

`platform.report` metrics for a lambda invocation are reported in the future invocation (in most cases the next one). Due to this, we will not have the report metrics for the last invocation till shutdown, and, as per the behavior of the extension prior to this PR, we will end up dropping the last metric. For periodic (hourly/daily/weekly) invocations, this will lead to no `platform.report` metric.

## Solution

During shutdown the extension gets a deadline of 2 seconds, this PR uses some of the 2 seconds to wait for the logs API to send us the` platform.report` metric for the last seen invocation. This wait is executed for each execution env/instance that served an invocation (NOT for each invocation). Per our initial benchmarks, the wait lasts for a maximum of 40ms (avg: ~5ms).

While the PR fixes the `platform.report` metric for successful invocations, for function crashes (timeouts/OOMs) it is still possible to miss the last `platform.report`.

Note that the `platform.report` metric for the last invocation can take as much as 45 minutes to be reported since they will be collected when the lambda execution env shuts down.

## Related Issues

Related to #334

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -24,7 +24,8 @@ https://github.com/elastic/apm-aws-lambda/compare/v1.2.0...main[View commits]
 
 [float]
 ===== Features
-- Create proxy transaction with error results if not reported by agent {lambda-pull}315[315]
+- experimental:[] Create proxy transaction with error results if not reported by agent {lambda-pull}315[315]
+- Wait for the final platform report metrics on shutdown {lambda-pull}347[347]
 
 [float]
 [[lambda-1.2.0]]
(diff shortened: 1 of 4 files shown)


AlexanderWert:
@lahsivjar One question: In case that the metrics get reported much later (e.g. with the shutdown, 45 mins after the last invocation), which timestamp is reported for the corresponding metric? Is it the timestamp of the previous invocation or the timestamp of when the `platform.report` event is processed (which is in that case 45min later)?

In the later case, would there be an easy way to use the invocation timestamp?

AlexanderWert:
I think I found the answer to my question  :-)

AlexanderWert (approved):

AlexanderWert:
looking forward to test this with my functions that run periodically once a day (right now I don't see any metrics at all, hope this will change it ) :-)

kruskall:
Somehow the assignee only got applied to the other extension PR.

I tested this on 8.6.0 and it worked fine. I discovered an issue with how we were compiling the test function and opened a PR to fix that (#350).

PR #355, opened 2023-01-01
build: Tag CSP resources according to policy

kruskall:
Reuse apm-server tags module to share labels for the apm-server team. Tag resource accordingly.

Related to https://github.com/elastic/apm-server/issues/9699

### DIFF
--- a/testing/benchmarking/main.tf
+++ b/testing/benchmarking/main.tf
@@ -33,6 +33,10 @@ provider "aws" {
   region = var.aws_region
 }
 
+module "tags" {
+  source  = "github.com/elastic/apm-server//testing/infra/terraform/modules/tags?depth=1"
+  project = "lambda-extension-benchmarks"
+}
 
 module "ec_deployment" {
   source = "github.com/elastic/apm-server/testing/infra/terraform/modules/ec_deployment"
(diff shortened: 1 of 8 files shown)


testing/benchmarking/main.tf
```diff
@@ -33,6 +33,10 @@ provider "aws" {
   region = var.aws_region
```
marclop, line 33: All the AWS resources will be tagged if you specify a `default_tags` field: https://registry.terraform.io/providers/hashicorp/aws/latest/docs#default_tags

kruskall: Do you mean instead of passing the tags to the tf modules ? I don't think we can assume they are using aws

kruskall:
@marclop Thanks for the help and discussion with this!

I've pushed some changes

tf/variables.tf
```diff
@@ -4,6 +4,11 @@ variable "aws_region" {
   default     = "eu-central-1"
 }
 
+variable "user_name" {
+  description = "Required username to use for prefixes"
+  type        = string
+}
+
```
marclop, line 11: Should this variable be also removed? It doesn't seem to be used.

marclop (approved):
Looks good!

tf/variables.tf
kruskall: I think it's used here: https://github.com/elastic/apm-aws-lambda/pull/355/files#diff-a2fd895ac25407846fd26b6d4fc9a4482f8cab6767431f76535be5ab0abe15eeR7 to set the project field.

marclop: never mind then!

PR #356, opened 2023-01-01
build: update notice to 2023

kruskall:
It's 2023 :tada: 

CI is failing because the year is not up to date.

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -1,4 +1,4 @@
-Copyright 2022 Elasticsearch BV
+Copyright 2023 Elasticsearch BV
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.


marclop (approved):

NOTICE.txt
```diff
@@ -1,4 +1,4 @@
-Copyright 2022 Elasticsearch BV
+Copyright 2023 Elasticsearch BV
```
marclop, line 1: Shouldn't it be `2022-2023`?

```suggestion
Copyright 2022-2023 Elasticsearch BV
```

kruskall: Nope, the template says `currentYear`: https://github.com/elastic/apm-aws-lambda/blob/main/scripts/templates/NOTICE.txt.tmpl#L13

kruskall: I'm not sure what we're doing in other repositories. It might be worth to look into it and make it more consistent

marclop: The APM Server used that format, that's why I was suggesting we use that format here as well.

kruskall: I've looked at a few repositories and I think the lambda extension is the only one using `currentYear`. I'll gather more data and open a followup issue/PR.

PR #358, opened 2023-01-10
Process platform report metrics when extn is lagging

lahsivjar:
The current extension will drop the `platform.report` logs if the platform report metrics are reported in an invocation that does not immediately succeed in the invocation generating the platform report. To prevent the logs from getting dropped the PR does the following changes:
1. Updates the lifecycle of the cached invocation data to be cleared on receiving `platform.report` metric for its corresponding request ID.
2. Query the cached invocation data instead of using the previous event and drop the log event if the prev event doesn't match the event's request ID.

Part of #334


lahsivjar:
Apologies for the push after marking it ready for review. I realized there was an edge case when on shutdown there might be more than one `platform.report` pending. I have fixed this now.

axw (approved):
Looks good!

PR #362, opened 2023-01-17
Jenkins 2 GH Actions Workflow Migration

reakaleek:
## Details

While migrating to GH Actions I had to utilise GoReleaser in conjunction with buildx to be able to build ARM compatible artifacts.

## How to test

I ran following steps locally:

Run without pushing or publishing to AWS:
```shell
export BUILD_DATE=202301251438.25
export DOCKER_REGISTRY=fakeregistry
export DOCKER_IMAGE_NAME=test
export GORELEASER_CURRENT_TAG=v0.0.9999
go run github.com/goreleaser/goreleaser@v1.14.1 release --snapshot --rm-dist
```

Test aws publish script:

> ⚠️ You also have to export AWS_* environments for a non production account.

```shell
export BUILD_DATE=202301251120.04
export GOOS=linux
export GOARCH=arm64
export ELASTIC_LAYER_NAME=reakaleek-test 
export ARCHITECTURE=arm64
export VERSION=v0.0.9999
./.ci/publish-aws.sh
```

Test the GitHub release:

> ⚠️ But you have to modify .ci/release-github.sh and add the `--draft` option to the `gh release` command.

```shell
export VERSION=v0.0.9999
./.ci/release-github.sh
```

## Related Issues
- Part of: https://github.com/elastic/observability-robots/issues/1517
- Depends On: https://github.com/elastic/apm-aws-lambda/pull/370


apmproxy/receiver_test.go
```diff
@@ -68,7 +68,7 @@ func TestInfoProxy(t *testing.T) {
 	}()
 
 	hosts, _ := net.LookupHost("localhost")
-	url := "http://" + hosts[0] + ":1234"
+	url := "http://" + hosts[1] + ":1234"
```
reakaleek, line 71: `hosts[0]` yields the IPV6 address. However, due to the fact that IPV6 is not supported on the `ubuntu-*` runners (See: https://github.com/actions/runner-images/issues/668#issuecomment-624080758) the tests are failing.

Using `hosts[1]` instead yields the IPV4 address and fixes the test.

apmproxy/receiver_test.go
```diff
@@ -195,6 +195,8 @@ func Test_handleInfoRequest(t *testing.T) {
 		req.Header.Add(name, value)
 	}
 
+	time.Sleep(5 * time.Second)
```
reakaleek, line 198: Adding some sleep here made the test more stable. It was flaky under ubuntu runners.

.goreleaser.yaml
```diff
@@ -0,0 +1,81 @@
+project_name: apm-lambda-extension
```
reakaleek, line 1: Migrated to GoReleaser because we don't have ARM runners available in GH Actions workflows by default.

Dockerfile
```diff
@@ -3,6 +3,4 @@
 # https://github.com/docker-library/repo-info/tree/master/repos/alpine/remote
 FROM alpine@sha256:bc41182d7ef5ffc53a40b044e725193bc10142a1243f395ee852a8d9730fc2ad
 ARG EXTENSION_FILE
-COPY ${EXTENSION_FILE} /opt/elastic-apm-extension
```
v1v, line 6: This might cause some issues if `EXTENSION_FILE` != `elastic-apm-extension`, I don't know if `EXTENSION_FILE` will be always `elastic-apm-extension` but since it's an ARG I might say this line will be needed

.github/workflows/release.yml
```diff
@@ -0,0 +1,60 @@
...
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      - uses: elastic/apm-pipeline-library/.github/actions/docker-login@current
+        with:
+          registry: docker.elastic.co
+          secret: secret/observability-team/ci/docker-registry/prod
+          url: ${{ secrets.VAULT_ADDR }}
+          roleId: ${{ secrets.VAULT_ROLE_ID }}
+          secretId: ${{ secrets.VAULT_SECRET_ID }}
+      - uses: hashicorp/vault-action@v2
```
v1v, line 42: what do you think to use the same version we use in the apm-pipeline-library? `hashicorp/vault-action@v2.4.2` see https://github.com/elastic/apm-pipeline-library/blob/main/.github/actions/setup-vault-cli/action.yml#LL20C15-L20C44

.ci/release-github.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+set -euo pipefail
```
v1v, line 2: what do you think to add a brief comment about what's the goal of this script?

.ci/publish-aws.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
+set -euo pipefail
+
+export AWS_FOLDER=${AWS_FOLDER:-.aws}
+export SUFFIX_ARN_FILE=${SUFFIX_ARN_FILE:-arn-file.md}
+
+GOOS=${GOOS:?Pleas provide GOOS environment variable.}
```
v1v, line 7:
```suggestion
GOOS=${GOOS:?Please provide GOOS environment variable.}
```

.ci/publish-aws.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
```
v1v, line 1: nit:

```suggestion
#!/usr/bin/env bash
```

.ci/publish-aws.sh
```diff
@@ -0,0 +1,59 @@
...
+set -euo pipefail
```
v1v, line 2: What do you think a header comment about the aim of this script?

.goreleaser.yaml
```diff
@@ -0,0 +1,81 @@
...
+
+before:
```
v1v, line 3: It changes the make goals that likely are used by the dev team, can you clarify whether this new way of using `goreleaser` is good for them when running the licenser validations?

.github/workflows/test.yml
```diff
@@ -0,0 +1,61 @@
...
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v3
+        with:
+          go-version-file: 'go.mod'
+      - name: Build
+        run: make build
+      - name: Test
+        run: |
+          go install gotest.tools/gotestsum@latest
+          gotestsum --format testname --junitfile ${{ matrix.platform }}-junit-report.xml
```
v1v, line 31: What do you think if those steps are added to the `test` Make goal?

```suggestion
        run: make test TEST_REPORT=${{ matrix.platform }}-junit-report.xml
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,61 @@
...
+      - name: Go vet
+        run: go vet
+  notice:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: notice
+        run: |
+          make NOTICE.txt
+          git diff --exit-code --quiet && exit 0 || echo "regenerate NOTICE.txt" && exit 1
```
v1v, line 58: what do you think a new make goal called `notice-ci` with those steps?

```suggestion
        run: make notice-ci
```

.goreleaser.yaml
```diff
@@ -0,0 +1,81 @@
...
+  - &default-docker-image
+    id: linux-amd64-image
+    use: buildx
+    goos: linux
+    goarch: amd64
+    image_templates:
+      - 'docker.elastic.co/observability/{{ .ProjectName }}-x86_64:{{ trimprefix .Tag "v" }}'
+      - "docker.elastic.co/observability/{{ .ProjectName }}-x86_64:latest"
+    build_flag_templates:
+      - "--platform=linux/amd64"
```
v1v, line 49: NIT: What do you think to leverage the build metadata with the opencontainers ones?

```
    - "--label=org.opencontainers.image.created={{.Date}}"
    - "--label=org.opencontainers.image.title={{.ProjectName}}"
    - "--label=org.opencontainers.image.revision={{.FullCommit}}"
    - "--label=org.opencontainers.image.version={{.Version}}"
```

See https://goreleaser.com/customization/docker/#applying-docker-build-flags

.goreleaser.yaml
```diff
@@ -0,0 +1,81 @@
...
+      - NOTICE.txt
+      - dependencies.asciidoc
+
+  - <<: *default-docker-image
+    id: linux-arm64-image
+    goarch: arm64
+    image_templates:
+      - 'docker.elastic.co/observability/{{ .ProjectName }}-arm64:{{ trimprefix .Tag "v" }}'
+      - "docker.elastic.co/observability/{{ .ProjectName }}-arm64:latest"
+    build_flag_templates:
```
v1v, line 61: ditto

.github/workflows/release.yml
```diff
@@ -0,0 +1,60 @@
+name: release
```
v1v, line 1: Can you add the `opentelemetry.yml` one? 🙏 

Similar to https://github.com/elastic/apm-pipeline-library/blob/main/.github/workflows/opentelemetry.yml

.github/workflows/release.yml
```diff
@@ -0,0 +1,60 @@
...
+permissions:
+  contents: write
+  pull-requests: read
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    outputs:
+      os: ${{ steps.artifacts.outputs.os }}
+      arch: ${{ steps.artifacts.outputs.arch }}
```
v1v, line 22: Where are these outputs used?

.github/workflows/release.yml
```diff
@@ -0,0 +1,60 @@
...
+          method: approle
+          roleId: ${{ secrets.VAULT_ROLE_ID }}
+          secretId: ${{ secrets.VAULT_SECRET_ID }}
+          secrets: |
+            secret/observability-team/ci/service-account/apm-aws-lambda access_key_id | AWS_ACCESS_KEY_ID ;
+            secret/observability-team/ci/service-account/apm-aws-lambda secret_access_key | AWS_SECRET_ACCESS_KEY
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v4
+        with:
+          args: release --rm-dist
```
v1v, line 54: What do you think to use `make release` rather than using the goreleaser action? By doing so, there is a need to install go-releaser as part of the `make release` goal, and might help with removing complexity in the CI by using a standalone make goal

```suggestion
        run: make release
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,60 @@
...
+      - name: GitHub Release
+        run: make release-notes
+        env:
+          VERSION: ${{ github.ref_name }}
+
+
```
v1v, line 60: As done in https://github.com/elastic/apm-aws-lambda/pull/362/files#diff-ef8b2f5e4d21d81c0eba4754dea3a3dfc2c67d26284f195165b650cf4aff7beaL216-L226 what do yo think to notify the release status with a slack notification? See https://github.com/elastic/apm-pipeline-library/blob/main/.github/actions/slack-message/action.yml

v1v (commented):
Great work!! A few comments

.github/workflows/test.yml
```diff
@@ -0,0 +1,61 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v3
+        with:
+          go-version-file: 'go.mod'
+      - name: lint-prep
+        run: go mod tidy && git diff --exit-code
+      - name: lint-run
+        run: make lint
+      - name: Go vet
```
kruskall, line 49: **todo**: opened an https://github.com/elastic/apm-aws-lambda/issues/367 for this. We should probably make sure go vet is run as part of the lint task but we can iterate on it on a separate PR.

.goreleaser.yaml
kruskall, line 1: **issue**: what happened to reproducible builds ? AFAICS this also does not support `SOURCE_DATE_EPOCH`. Two identical commits should produce the same build output regardless of time or host the build was started on.
Be it a docker image or a zip archive.

kruskall (changes requested):
Left a comment about reproducible builds

apmproxy/receiver_test.go
kruskall: Thanks, I've opened https://github.com/elastic/apm-aws-lambda/issues/368 to keep track of it

.github/workflows/release.yml
reakaleek: Ah, yes.. thank you.. missed that. https://github.com/elastic/apm-aws-lambda/pull/362/commits/fbd2a6901714cf0c13a10e4c3708ca7f0b741bcb

.goreleaser.yaml
reakaleek: Thank you. This go lost in translation. I will try to replicate this behaviour.

.ci/publish-aws.sh
reakaleek: https://github.com/elastic/apm-aws-lambda/pull/362/commits/03751bc415ec09f7cf715f8aa753d1fcfe5f3636

.ci/publish-aws.sh
reakaleek: https://github.com/elastic/apm-aws-lambda/pull/362/commits/b564627bd718c5cb467d92375870fc8cbe516c50

.github/workflows/release.yml
reakaleek: https://github.com/elastic/apm-aws-lambda/pull/362/commits/0a7eefd81219f77b29abb5830a718ef7f4a471aa

.github/workflows/release.yml
reakaleek: https://github.com/elastic/apm-aws-lambda/pull/362/commits/88f138d4cfddba7d546cd88502220949eced7a0c

.github/workflows/test.yml
reakaleek: https://github.com/elastic/apm-aws-lambda/pull/362/commits/37f2b84fc08fc9e4dacb7d8fdc6afe114cc41288

.github/workflows/release.yml
reakaleek: https://github.com/elastic/apm-aws-lambda/pull/362/commits/7b1c80c4bf936c08d00e59c0e8c8ad5096e93ded

.github/workflows/release.yml
reakaleek: https://github.com/elastic/apm-aws-lambda/pull/362/commits/b23272aaa3dcea334526b36ca458d0406d31f04b

.goreleaser.yaml
reakaleek: https://github.com/elastic/apm-aws-lambda/pull/362/commits/f3b775a20760b2848b539adcc00cfd9b85bf4f08

.goreleaser.yaml
reakaleek: https://github.com/elastic/apm-aws-lambda/pull/362/commits/f3b775a20760b2848b539adcc00cfd9b85bf4f08

Dockerfile
reakaleek: https://github.com/elastic/apm-aws-lambda/pull/362/commits/38753055563a324cf0231c74c3d80506e311e30a

.ci/release-github.sh
reakaleek: https://github.com/elastic/apm-aws-lambda/pull/362/commits/26ad58d73304cd8bdd2270f2f250f51b1076c10c

.github/workflows/test.yml
reakaleek: https://github.com/elastic/apm-aws-lambda/pull/362/commits/6b3bb5aca734423a653af535bca4d72799af510d

.goreleaser.yaml
reakaleek: Moved several scripts to Makefile rules

https://github.com/elastic/apm-aws-lambda/pull/362/commits/02aa3fe7c40693271994b80ee86e89a0585f180e
https://github.com/elastic/apm-aws-lambda/pull/362/commits/b5323fa5cadad6531ffa956875457419abe8f28c
https://github.com/elastic/apm-aws-lambda/pull/362/commits/6b3bb5aca734423a653af535bca4d72799af510d
https://github.com/elastic/apm-aws-lambda/pull/362/commits/e527f43daf19eeb30cfd25dbad82ee9f39fa3987
https://github.com/elastic/apm-aws-lambda/pull/362/commits/818ba189182d8b9f02416284e1d75d785068b3d9

reakaleek: @kruskall can you please also check this and suggest any adjustments to the Makefile if necessary? Thank you!

.goreleaser.yaml
reakaleek: For my understanding:

to make the build reproducible, I have to touch the artifacts with the timestamp of the git log, right?
Is there anything else I am missing?

reakaleek: This is my attempt, let me know if this works!

https://github.com/elastic/apm-aws-lambda/pull/362/commits/1c33f8f812663035211c1cac4251c3c284da3637

cc @v1v

.ci/publish-aws.sh
```diff
@@ -0,0 +1,74 @@
...
+
+export AWS_FOLDER=${AWS_FOLDER:-.aws}
+export SUFFIX_ARN_FILE=${SUFFIX_ARN_FILE:-arn-file.md}
+
+GOOS=${GOOS:?Please provide GOOS environment variable.}
+GOARCH=${GOARCH:?Please provide GOARCH environment variable.}
+ELASTIC_LAYER_NAME=${ELASTIC_LAYER_NAME:?Please provide ELASTIC_LAYER_NAME environment variable.}
+ARCHITECTURE=${ARCHITECTURE:?Please provide ARCHITECTURE environment variable.}
+VERSION=${VERSION:?Please provide VERSION environment variable. e.g. current git tag}
+
```
v1v, line 17:
```suggestion
BUILD_DATE=${BUILD_DATE:?Please provide BUILD_DATE environment variable.}

```

v1v (commented):
LGTM

Question regarding testing this PR locally, would you mind adding what command did you run locally? 🙇

.goreleaser.yaml
```diff
@@ -0,0 +1,86 @@
...
+
+archives:
+  - id: zip
+    format: zip
+    name_template: |-
+      {{ .Tag }}-{{ .Os }}-{{ .Arch }}
+    files:
+      - NOTICE.txt
+      - dependencies.asciidoc
+    rlcp: true # https://goreleaser.com/deprecations/#archivesrlcp
```
v1v, line 31: This option was not available in my current gorelease version:

```
   • releasing...     
   • loading config file       file=.goreleaser.yaml
   ⨯ release failed after 0.00s error=yaml: unmarshal errors:
  line 31: field rlcp not found in type config.Archive
make: *** [dist] Error 1
```

I've just updated the version with `go install github.com/goreleaser/goreleaser@latest`

I wonder if the version should be pinned in the makefile, see my proposal

Makefile
```diff
@@ -12,183 +9,58 @@ DATE_FMT = +%Y%m%d%H%M.%S
...
-GO_BUILDFLAGS ?= -trimpath -ldflags="-s -w"
+clean:
+	@rm -rf dist/
+	@docker image ls "$(DOCKER_REGISTRY)/$(DOCKER_IMAGE_NAME)*" -aq | xargs docker rmi --force
 
-ifndef GOARCH
-	GOARCH=amd64
-endif
+dist:
+	@goreleaser release --snapshot --rm-dist
```
v1v, line 23:
```suggestion
	@go run github.com/goreleaser/goreleaser@v1.14.1 release --snapshot --rm-dist
```

Makefile
```diff
@@ -12,183 +9,58 @@ DATE_FMT = +%Y%m%d%H%M.%S
...
 
-# Transform GOARCH into the architecture of the extension layer
-ifeq ($(GOARCH),amd64)
-	ARCHITECTURE=x86_64
-else
-	ARCHITECTURE=arm64
-endif
+.PHONY: release
+release:
+	@goreleaser release --rm-dist
```
v1v, line 27:
```suggestion
	@go run github.com/goreleaser/goreleaser@v1.14.1 release --rm-dist
```

Makefile
reakaleek: Ah nice, I was not aware of this. Even though, I saw it in some other commands.

https://github.com/elastic/apm-aws-lambda/pull/362/commits/8bd2e4302b57a015a963ec39b16642bc946df466

Makefile
reakaleek: https://github.com/elastic/apm-aws-lambda/pull/362/commits/8bd2e4302b57a015a963ec39b16642bc946df466

.goreleaser.yaml
reakaleek: https://github.com/elastic/apm-aws-lambda/pull/362/commits/8bd2e4302b57a015a963ec39b16642bc946df466

.ci/publish-aws.sh
reakaleek: https://github.com/elastic/apm-aws-lambda/pull/362/commits/ec50ef8bdece29b6d5b5b7461a2021a3f4c04d64

reakaleek:
> LGTM
> 
> Question regarding testing this PR locally, would you mind adding what command did you run locally? 🙇

Updated the description: https://github.com/elastic/apm-aws-lambda/pull/362#issue-1536371273

v1v (approved):
💯 super great work! Thanks

.ci/publish-aws.sh
```diff
@@ -0,0 +1,76 @@
...
+(cd "${tmp_dir}" && zip -X -r "../../${new_zip_file}" .)
+
+for region in $ALL_AWS_REGIONS; do
+  echo "Publish ${FULL_LAYER_NAME} in ${region}"
+  publish_output=$(aws lambda \
+    --output json \
+    publish-layer-version \
+    --region="${region}" \
+    --layer-name="${FULL_LAYER_NAME}" \
+    --compatible-architectures="${ARCHITECTURE}" \
```
reakaleek, line 57: Added this so the compatible architecture is visible in the UI.

image

Makefile
```diff
@@ -12,183 +9,54 @@ DATE_FMT = +%Y%m%d%H%M.%S
...
-	$(error AWS_ACCESS_KEY_ID is undefined)
-endif
-ifndef AWS_SECRET_ACCESS_KEY
-	$(error AWS_SECRET_ACCESS_KEY is undefined)
-endif
-	GOARCH=${GOARCH} make build
-	GOARCH=${GOARCH} make zip
-	$(MAKE) publish
-
-zip: build
```
kruskall, line 68: **issue**: I think these tasks cannot be removed, we need a zip/build target to quickly build the extension and use the terraform script in `tf` for testing.

Dockerfile
```diff
@@ -3,6 +3,10 @@
 # https://github.com/docker-library/repo-info/tree/master/repos/alpine/remote
 FROM alpine@sha256:bc41182d7ef5ffc53a40b044e725193bc10142a1243f395ee852a8d9730fc2ad
 ARG EXTENSION_FILE
+ARG BUILD_DATE
```
kruskall, line 6: **issue**: I don't think this is right. The dockerfile is only copying the build artifacts to the docker image. There should be no need to change the file time inside the docker image as the build artifacts should have already been built reproducibly.

.goreleaser.yaml
kruskall: I left a few comments, some of the tasks are also used for development/testintg. 

Apologies, I think I'm missing some context here, is there an issue with the current makefile ? I see the building logic has been moved to goreleaser but is there any special reason behind it ?

kruskall (changes requested):

.goreleaser.yaml
kruskall: @reakaleek 

I left a few comments, some of the tasks are also used for development/testintg.

Apologies, I think I'm missing some context here, is there an issue with the current makefile ? I see the building logic has been moved to goreleaser but is there any special reason behind it ?

Dockerfile
reakaleek: The reason for this way of doing it is a "limitation" of GoReleaser as it copies the artifacts and files used in the Dockerfile to a temporary directory and builds it from there.

You can see this behavior when running `make dist` 

image

Makefile
reakaleek: I'm currently working on the smoke tests aswell and there I re-added a build rule. (See: https://github.com/elastic/apm-aws-lambda/pull/370)

You can still create the zip using `make dist`. There are ways to skip unnecessary pipes like `--skip-docker`.

Makefile
```diff
@@ -12,183 +9,54 @@ DATE_FMT = +%Y%m%d%H%M.%S
...
-ifndef ELASTIC_LAYER_NAME
-	$(error ELASTIC_LAYER_NAME is undefined)
-endif
-
-validate-aws-default-region:
-ifndef AWS_DEFAULT_REGION
-	$(error AWS_DEFAULT_REGION is undefined)
-endif
+.PHONY: check-extension
+check-extension:
```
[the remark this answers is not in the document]
reakaleek: https://github.com/elastic/apm-aws-lambda/pull/362/commits/4a72e87cb308bafd95d7133d0a179546b11ec419

.goreleaser.yaml
reakaleek: The main reason is that the current scripts were created with running on ARM CI runners in mind. Hence, in Jenkins we would just run the whole script on an ARM instance in parallel to the AMD build.

However, in GH Actions, we don't have ARM runners available by default, unless we would self-host them. Which is currently not an option. (Maybe in the future).

Therefore, we decided to use GoReleaser as it easens the multi-architecture build of artifacts.

.goreleaser.yaml
```diff
@@ -0,0 +1,85 @@
...
+    build_flag_templates:
+      - "--platform=linux/arm64"
+      - "--build-arg=EXTENSION_FILE={{ .ProjectName }}"
+      - "--build-arg=BUILD_DATE={{ .Env.BUILD_DATE }}"
+      - "--label=org.opencontainers.image.created={{ .Date }}"
+      - "--label=org.opencontainers.image.title={{ .ProjectName }}"
+      - "--label=org.opencontainers.image.revision={{ .FullCommit }}"
+      - "--label=org.opencontainers.image.version={{ .Version }}"
+
+publishers:
```
kruskall, line 69: **question**: will this always publish to aws when running make tasks ?

Makefile
```diff
@@ -12,183 +9,50 @@ DATE_FMT = +%Y%m%d%H%M.%S
...
 # 1. 'date -d': Busybox and GNU coreutils.
 # 2. 'date -r': BSD date. It does not support '-d'.
-BUILD_DATE = $(shell date -u -d "@${SOURCE_DATE_EPOCH}" "${DATE_FMT}" 2>/dev/null || date -u -r "${SOURCE_DATE_EPOCH}" "${DATE_FMT}")
+export BUILD_DATE = $(shell date -u -d "@${SOURCE_DATE_EPOCH}" "${DATE_FMT}" 2>/dev/null || date -u -r "${SOURCE_DATE_EPOCH}" "${DATE_FMT}")
+
 
-GO_BUILDFLAGS ?= -trimpath -ldflags="-s -w"
+clean:
+	@rm -rf dist/
+	@docker image ls "$(DOCKER_REGISTRY)/$(DOCKER_IMAGE_NAME)*" -aq | xargs docker rmi --force
```
kruskall, line 17: **issue**: this will fail if there is no docker image

.goreleaser.yaml
kruskall: Nope, because we're using `--snapshot`

kruskall (commented):
Just a minor comment

Dockerfile
```diff
@@ -3,6 +3,10 @@
...
 COPY ${EXTENSION_FILE} /opt/elastic-apm-extension
-COPY /NOTICE.txt /opt/NOTICE.txt
-COPY /dependencies.asciidoc /opt/dependencies.asciidoc
+COPY NOTICE.txt dependencies.asciidoc /opt/
+
+# Related to reproducible builds
+RUN find /opt -exec touch -t "${BUILD_TIME}" {} \;
```
kruskall, line 11: **issue**: did you mean `BUILD_DATE` ?

kruskall (changes requested):

Dockerfile
kruskall: I tried running the command and got the following:

```
release failed after 1s                  error=docker build failed: failed to build docker.elastic.co/observability/apm-lambda-extension-arm64:1.2.0: exit status 125: unknown flag: --builder
```

it seems goreleaser is using `docker-buildx` under the hood. Installing `docker-buildx` fixed the issues. (should be documented, opened a followup issue: https://github.com/elastic/apm-aws-lambda/issues/372)

building the docker image is failing but I think that's related to a typo. I've left a comment below.

Makefile
```diff
@@ -12,183 +9,50 @@ DATE_FMT = +%Y%m%d%H%M.%S
...
 
-ifndef GOARCH
-	GOARCH=amd64
-endif
+dist:
+	@go run github.com/goreleaser/goreleaser@v1.14.1 release --snapshot --rm-dist
```
kruskall, line 20: **suggestion**: can we skip building the docker images and move that to separate task ? the docker version requires a new dependency and it's not really used for testing/development

.goreleaser.yaml
```diff
@@ -0,0 +1,85 @@
...
+    goarch:
+      - amd64
+      - arm64
+    flags:
+      - -trimpath
+    ldflags:
+      - -s -w
+    binary: "extensions/{{ .ProjectName }}"
+
+archives:
```
kruskall, line 22: **question**: is this the spec for this task ? https://goreleaser.com/customization/archive/

Can't you use the `mtime` fields and get rid of the touch workaround ? Or am I missing something ?

reakaleek: Oh nice, thank you for that. I was not aware that `mtime` could be used.

reakaleek: https://github.com/elastic/apm-aws-lambda/pull/362/commits/8975604b5c891a79e4c925b6aadfed9d9c4b53bb

Dockerfile
reakaleek: thank you for noticing: https://github.com/elastic/apm-aws-lambda/pull/362/commits/d15a6b9839f1db730b05ded25e0d0a026f0d0636

Makefile
```diff
@@ -1,194 +1,59 @@
...
 # See https://reproducible-builds.org/specs/source-date-epoch/
 SOURCE_DATE_EPOCH ?= $(shell git log -1 --pretty=%ct)
 DATE_FMT = +%Y%m%d%H%M.%S
+DATE_FMT_RFC_3339 = +%Y-%m-%dT%H:%M:%SZ
 # Fallback mechanism to support other systems:
 # 1. 'date -d': Busybox and GNU coreutils.
 # 2. 'date -r': BSD date. It does not support '-d'.
-BUILD_DATE = $(shell date -u -d "@${SOURCE_DATE_EPOCH}" "${DATE_FMT}" 2>/dev/null || date -u -r "${SOURCE_DATE_EPOCH}" "${DATE_FMT}")
+export BUILD_DATE = $(shell date -u -d "@${SOURCE_DATE_EPOCH}" "${DATE_FMT}" 2>/dev/null || date -u -r "${SOURCE_DATE_EPOCH}" "${DATE_FMT}")
+export BUILD_DATE_RFC_3339 = $(shell date -u -d "@${SOURCE_DATE_EPOCH}" "${DATE_FMT_RFC_3339}" 2>/dev/null || date -u -r "${SOURCE_DATE_EPOCH}" "${DATE_FMT_RFC_3339}")
```
kruskall, line 14: **issue**: `$(shell ...)` is a make variable. If you convert these to env variable you need to remove the `shell`. This should fix the error:

```
date: '': No such file or directory
date: '': No such file or directory
date: '': No such file or directory
date: '': No such file or directory
```

.goreleaser.yaml
```diff
@@ -0,0 +1,87 @@
...
+  - id: zip
+    format: zip
+    name_template: |-
+      {{ .Tag }}-{{ .Os }}-{{ .Arch }}
+    files:
```
kruskall, line 27: **issue**: I think you also need to apply mtime to files. `builds_info` only applies to binaries afaics.

kruskall (changes requested):

Makefile
reakaleek: I actually got rid of this and used the goreleaser variable `.CommitTimestamp` instead.

https://github.com/elastic/apm-aws-lambda/pull/362/commits/ccb7116483e31cb1ffac3080a25964ebe4b84f95

.goreleaser.yaml
reakaleek: https://github.com/elastic/apm-aws-lambda/pull/362/commits/645fcfddc05f72acb45db59c1e6f20b4aa18924f

Makefile
reakaleek: I tried this:

```shell
make clean
make clean
echo $?
0
```

and I get the exit code 0. can you elaborate? maybe I'm missing something.

Makefile
reakaleek: https://github.com/elastic/apm-aws-lambda/pull/362/commits/2c5201016e08901ffb500210e01e7ef94b4cbc13

I re-added the `zip` rule, hope this works aswell.

Makefile
kruskall: Note: that means losing support for `SOURCE_DATE_EPOCH` but it should be fine since it's not mandatory and the commit timestamp should not change

Makefile
kruskall: Thanks!

Makefile
kruskall: mmh

```
$ make clean 
"docker rmi" requires at least 1 argument.
See 'docker rmi --help'.

Usage:  docker rmi [OPTIONS] IMAGE [IMAGE...]

Remove one or more images
make: *** [Makefile:11: clean] Error 123
```

Are you perhaps on macOS ? FreeBSD xargs ignores empty args if I recall correctly

reakaleek: Oh, yes I'm on mac. Let me try to figure out on how to fix this.

Dockerfile
kruskall: Ok, I see some progress but the build is still failing:

```
#8 [4/4] RUN find /opt -exec touch -amdt "1675338235" {} ;
#8 0.240 exec /bin/sh: exec format error
#8 ERROR: executor failed running [/bin/sh -c find /opt -exec touch -amdt "${COMMIT_TIMESTAMP}" {} \;]: exit code: 1
------
 > [4/4] RUN find /opt -exec touch -amdt "1675338235" {} ;:
#8 0.240 exec /bin/sh: exec format error
------
ERROR: failed to solve: executor failed running [/bin/sh -c find /opt -exec touch -amdt "${COMMIT_TIMESTAMP}" {} \;]: exit code: 1
```

Makefile
reakaleek: Alright, this should do the trick: https://github.com/elastic/apm-aws-lambda/pull/362/commits/d1b6db3aa0903759547e4d042d951c28e4aba315

kruskall: Works like a charm!

Dockerfile
kruskall: @reakaleek I think this is the last issue. Is the docker build working for you ?

reakaleek: Interesting. Yes, it works for me. (Apple macOS M1)

And this [workflow](https://github.com/elastic/apm-aws-lambda/actions/runs/4076401474/jobs/7024085307) shows that it should work with Ubuntu on amd64.

Looking at your error message, if I recall correctly, this could be a buildx setup issue.

kruskall: I was able to reproduce the issue by building the arm64 image.

`docker buildx build --platform linux/arm64 .`

I'm on linux (amd64). I'll investigate more

Dockerfile
```diff
@@ -3,6 +3,10 @@
 # https://github.com/docker-library/repo-info/tree/master/repos/alpine/remote
 FROM alpine@sha256:bc41182d7ef5ffc53a40b044e725193bc10142a1243f395ee852a8d9730fc2ad
```
kruskall, line 4: We need to add `--platform` for this to work and make sure we are using the main image instead of the arch-specific one.

According to https://github.com/docker-library/repo-info/tree/master/repos/alpine/remote the latest 3.16.x is 3.16.3


```suggestion
FROM --platform=$BUILDPLATFORM alpine@sha256:b95359c2505145f16c6aa384f9cc74eeff78eb36d308ca4fd902eeeb0a0b161b
```

Dockerfile
```diff
@@ -3,6 +3,10 @@
...
 ARG EXTENSION_FILE
+ARG COMMIT_TIMESTAMP
 COPY ${EXTENSION_FILE} /opt/elastic-apm-extension
-COPY /NOTICE.txt /opt/NOTICE.txt
-COPY /dependencies.asciidoc /opt/dependencies.asciidoc
+COPY NOTICE.txt dependencies.asciidoc /opt/
+
+# Related to reproducible builds
+RUN find /opt -exec touch -amdt "${COMMIT_TIMESTAMP}" {} \;
```
kruskall, line 11: This is silently failing.

Adding `--debug` to the gorelease command in the make task shows:

```
    • #8 0.264 touch: invalid date 't'               cmd=[docker buildx] cwd=dist/goreleaserdocker3058620621
    • #8 0.265 touch: invalid date 't'               cmd=[docker buildx] cwd=dist/goreleaserdocker3058620621
    • #8 0.265 touch: invalid date 't'               cmd=[docker buildx] cwd=dist/goreleaserdocker3058620621
    • #8 0.266 touch: invalid date 't'               cmd=[docker buildx] cwd=dist/goreleaserdocker3058620621
```

Opening a shell into an alpine container: 

```
$ docker run -it alpine:3.16.3 /bin/sh
/ # touch -h
BusyBox v1.35.0 multi-call binary.

Usage: touch [-cham] [-d DATE] [-t DATE] [-r FILE] FILE...

Update mtime of FILEs

        -c      Don't create files
        -h      Don't follow links
        -a      Change only atime
        -m      Change only mtime
        -d DT   Date/time to use
        -t DT   Date/time to use
        -r FILE Use FILE's date/time
```

The correct command should be `RUN find /opt -exec touch -am -d "${COMMIT_TIMESTAMP}" -t "${COMMIT_TIMESTAMP}" {} \;`

but then `touch` complains about the date format being invalid.

Previously we used `date` to create a posix compliant task (https://github.com/elastic/apm-aws-lambda/pull/362/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L11-L15) but since the command is run inside an alpine container we can assume busybox and tailor the syntax.

The final command should be something like:

```suggestion
RUN find /opt -exec touch -am -d $(date -u -d @"${COMMIT_TIMESTAMP}" "+%Y%m%d%H%M.%S") -t $(date -u -d @"${COMMIT_TIMESTAMP}" "+%Y%m%d%H%M.%S") {} \;
```

kruskall (changes requested):

Dockerfile
reakaleek: https://github.com/elastic/apm-aws-lambda/pull/362/commits/2a93275a70640d12ba76d8e80c3ebdbd18fe62a9

Dockerfile
reakaleek: Thank you very much for this.

reakaleek: https://github.com/elastic/apm-aws-lambda/pull/362/commits/daa0ceec202377c22d98238e4b84d67088b135e1

kruskall (approved):
LGTM!

PR #370, opened 2023-01-26
Smoketests 2 GH Actions Workflow Migration

reakaleek:
## Details

Migrate smoke-tests.groovy to GH Actions Workflow.

There is some loop logic that would run on a single item. For the sake of simplicity, I omitted this logic for now and just run it directly on the `./tf` folder.

## How to test

Adding the `on.pull_request` triggers the workflow.

## Related Issues
Issue: https://github.com/elastic/observability-robots/issues/1517
Depends on: https://github.com/elastic/apm-aws-lambda/pull/371


reakaleek:
You can find the last successful run here: https://github.com/elastic/apm-aws-lambda/actions/runs/4043341456/jobs/6952144043 (It was using the `on.pull_request` trigger`)

reakaleek:
@elastic/observablt-ci Even though the `on.workflow_dispatch` trigger is there, I cannot trigger it from the UI.

Is this only available if the workflow file is on the main trunk?

v1v:
> Is this only available if the workflow file is on the main trunk?

That's the behaviour

>> To trigger the workflow_dispatch event, your workflow must be in the default branch. 

As per https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow#configuring-a-workflow-to-run-manually

Although somehow it gets triggered for this particular PR -> https://github.com/elastic/apm-aws-lambda/actions/runs/4043635731/jobs/6952794328

Maybe I miss something but it should not be the case as documented in https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

.github/workflows/smoke-tests.yml
```diff
@@ -0,0 +1,73 @@
...
+      smoketest_versions:
+        description: Set SMOKETEST_VERSIONS environment variable
+        default: latest
+  push:
+    branches:
+      - main
+
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: true
```
v1v, line 15: I'm not sure if we want that behaviour, I'd say for PRs might make sense, but for merge commits, I'd prefer to keep them running, so every commit will have a smoke test ideally.

Maybe a timeout could be a good one to be added instead, what do you think?

tf/test.sh
```diff
@@ -15,17 +15,32 @@ echo "-> Creating the underlying infrastructure..."
 terraform init | tee tf.log
 terraform apply -auto-approve | tee -a tf.log
 
-AWS_REGION=$(terraform output -raw aws_region)
+# https://github.com/hashicorp/setup-terraform/issues/167#issuecomment-1090760365
+if [[ -z "${GITHUB_WORKFLOW}" ]]; then
```
v1v, line 19: what do you think by using a binary alias? See below 
```suggestion
TERRAFORM_WRAPPER=terraform-bin
if [[ -z "${GITHUB_WORKFLOW}" ]]; then
  TERRAFORM_WRAPPER=terraform
fi
AWS_REGION=$(${TERRAFORM_WRAPPER} output -raw aws_region)
```

tf/test.sh
```diff
@@ -15,17 +15,32 @@ echo "-> Creating the underlying infrastructure..."
...
+  ES_HOST=$(terraform output -raw elasticsearch_url)
+  ES_USER=$(terraform output -raw elasticsearch_username)
+  ES_PASS=$(terraform output -raw elasticsearch_password)
+else
+  ES_HOST=$(terraform-bin output -raw elasticsearch_url)
+  ES_USER=$(terraform-bin output -raw elasticsearch_username)
+  ES_PASS=$(terraform-bin output -raw elasticsearch_password)
+fi
```
v1v, line 41: then

```suggestion
ES_HOST=$(${TERRAFORM_WRAPPER} -bin output -raw elasticsearch_url)
ES_USER=$(${TERRAFORM_WRAPPER}  output -raw elasticsearch_username)
ES_PASS=$(${TERRAFORM_WRAPPER}  output -raw elasticsearch_password)
```

.github/workflows/smoke-tests.yml
reakaleek: I added this because the smoke tests cannot run in parallel and will fail with terraform resource conflicts.

v1v: Gotcha, what about removing the `cancel-in-progress`?

tf/test.sh
reakaleek: https://github.com/elastic/apm-aws-lambda/pull/370/commits/f4e7c7c6a67de0665372e12dfde7c6817ec2e4fd

tf/test.sh
reakaleek: https://github.com/elastic/apm-aws-lambda/pull/370/commits/f4e7c7c6a67de0665372e12dfde7c6817ec2e4fd

.github/workflows/smoke-tests.yml
reakaleek: Ah, so the latter job will wait in queue until the first one is finished. is this correct?

reakaleek: https://github.com/elastic/apm-aws-lambda/pull/370/commits/ba2035694af283162e34f70d9b05ee4a6525daba

v1v: That's my understanding as per [concurrency](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency)

reakaleek: > so every commit will have a smoke test ideally.

Also in Jenkins, this was running only when triggered manually.

Makefile
```diff
@@ -19,6 +19,9 @@ clean:
 dist:
 	@go run github.com/goreleaser/goreleaser@v1.14.1 release --snapshot --rm-dist
 
+build:
```
v1v, line 22: Unrelated to the changes? I cannot find this make build command

tf/main.tf
```diff
@@ -55,7 +55,7 @@ module "lambda_layer_local" {
   description         = "AWS Lambda Extension Layer for Elastic APM - smoke testing"
   compatible_runtimes = ["nodejs16.x"]
 
-  source_path = "../bin/"
+  source_path = "../dist/apm-lambda-extension_linux_amd64_v1/"
```
v1v, line 58: will this particular name change in the future? If so, what do you think adding a comment regarding where this name is coming from?

v1v (approved):
LGTM, with just a couple of minor comments

tf/main.tf
reakaleek: There is actually a comment about this in the GoRelease documentation.. will find and add it here.

But from my understanding, this will only change if changing the project name OR potentially when upgrading the goreleaser version.

Makefile
reakaleek: Not sure what you mean 😅

reakaleek: Ah, it's actually used here: https://github.com/elastic/apm-aws-lambda/blob/7b856d97ff778feea4e8d1a11ceb422a43a46b7b/Makefile#L72

v1v: Thanks, I was blind! 🙏

tf/main.tf
reakaleek: https://goreleaser.com/customization/build/#a-note-about-folder-names-inside-dist

> By default, GoReleaser will create your binaries inside dist/${BuildID}_${BuildTarget}, which is an unique directory per build target in the matrix.
>
> Those names have no guarantees of remaining the same from one version to another. If you really need to access them from outside GoReleaser, you should be able to consistently get the path of a binary by parsing dist/artifacts.json.

reakaleek: https://github.com/elastic/apm-aws-lambda/pull/370/commits/a1efdaf3034d5b5c9b759794c1123955a3754237

kruskall (approved):
LGTM!

.github/workflows/smoke-tests.yml
```diff
@@ -0,0 +1,72 @@
+name: smoke-tests
+
+on:
+  pull_request: ~
```
reakaleek, line 4:
```suggestion
```

PR #384, opened 2023-04-10
Allow metadata in register transaction

### LINKED ISSUE #352: Extension might drop data if the first call for an execution environment fails
**APM AWS Lambda extension version** (`apm-aws-lambda version`): All

**Describe the bug**

If the first call to an execution environment executing a lambda function (`coldstart: true`) fails due to timeout or crashes causing a reset of the execution env then the extension will drop any data it has accumulated. The drops happen due to agent not being able to report `metadata` to the extension. This issue will also lead to failures while reporting the proxy transaction that is introduced in issue #315 when the first invocation in the execution environment fails.

**To Reproduce**
1. Launch a lambda function that always times out, uses apm-agent instrumentation for lambda, and has the lambda-extension as a layer (can also use the [testing](https://github.com/elastic/apm-aws-lambda/tree/main/testing) setup, with modification to the lambda function so that it times out, to create the lambda and EC infrastructure).
4. Invoke the lambda function.
5. Observe that there are no events (traces, logs or metrics) available in Elastic cloud.
6. Observe the lambda cloudwatch logs to confirm the error log `Dropping lambda data due to error: metadata is not yet available`.

**Expected behavior**
Lambda extension is able to send events for cases when the function times out/crashes on its first invocation. Some possible solutions are:
1. Add a new API in the extension to accept metadata and make sure agents report metadata as soon as it is available.
2. Create proxy metadata in the extension based on whatever data the extension has available.

**Provide logs (if relevant)**:
[lambda_logs_for_first_invocation_timeout.csv](https://github.com/elastic/apm-aws-lambda/files/10216007/lambda_logs_for_first_invocation_timeout.csv)

lahsivjar:
The PR updates the `/register/transaction` endpoint to accept `NDJSON`. The `/register/transaction` endpoint can now accept `metadata` as the first event in `NDJSON` body and cache it for future calls. A complete list of changes introduced in the PR is as follows:

1. Change the `Content-Type` for the `/register/transaction` endpoint from `application/vnd.elastic.apm.transaction+json` to `application/vnd.elastic.apm.transaction+ndjson`.
2. Add support for `deflate` and `gzip` content encoding for the `/register/transaction` endpoint.
3. Allow `metadata` as the first event in the `NDJSON` body to the `/register/transaction` endpoint. The `metadata` is cached and used to send data to APM-Server. The cached metadata is NOT updated by the metadata sent via the intake endpoint.

### Related Issue

Closes #352

### DIFF
--- a/apmproxy/receiver.go
+++ b/apmproxy/receiver.go
@@ -29,10 +29,9 @@ import (
 	"time"
 
 	"github.com/elastic/apm-aws-lambda/accumulator"
-	"github.com/tidwall/gjson"
 )
 
-const txnRegistrationContentType = "application/vnd.elastic.apm.transaction+json"
+const txnRegistrationContentType = "application/vnd.elastic.apm.transaction+ndjson"
 
 // StartReceiver starts the server listening for APM agent data.
 func (c *Client) StartReceiver() error {
@@ -76,12 +75,12 @@ func (c *Client) Shutdown() error {
 // URL: http://server/
 func (c *Client) handleInfoRequest() (func(w http.ResponseWriter, r *http.Request), error) {
 	// Init reverse proxy
-	parsedApmServerUrl, err := url.Parse(c.serverURL)
+	parsedApmServerURL, err := url.Parse(c.serverURL)
 	if err != nil {
 		return nil, fmt.Errorf("could not parse APM server URL: %w", err)
 	}
 
-	reverseProxy := httputil.NewSingleHostReverseProxy(parsedApmServerUrl)
+	reverseProxy := httputil.NewSingleHostReverseProxy(parsedApmServerURL)
 
 	reverseProxy.Transport = c.client.Transport.(*http.Transport).Clone()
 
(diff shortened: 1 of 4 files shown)


basepi:
> The cached metadata is NOT updated by the metadata sent via the intake endpoint.

This is slightly concerning to me. We have a mechanism for adding "late-breaking" metadata in the python agent, for metadata that is only available at request time. For now, this should not be a problem -- the only late metadata we add is for lambda, and that metadata will be available when we send the partial transaction.

I just wanted to call it out as something we should explicitly state when we update the [spec](https://github.com/elastic/apm/blob/main/specs/agents/tracing-instrumentation-aws-lambda.md#data-flushing) with partial transaction support.

basepi (approved):
👍 to stated changes, haven't reviewed the code.

trentm (commented):
The functionality sounds good. I won't have a chance to test this for a little while yet (hopefully this week).

apmproxy/receiver.go
```diff
@@ -29,10 +29,9 @@ import (
 	"time"
 
 	"github.com/elastic/apm-aws-lambda/accumulator"
-	"github.com/tidwall/gjson"
 )
 
-const txnRegistrationContentType = "application/vnd.elastic.apm.transaction+json"
+const txnRegistrationContentType = "application/vnd.elastic.apm.transaction+ndjson"
```
marclop, line 34: is it OK to break backwards compatibility here? Is there a concern for using older agents? AFAICS, if older agents attempt to send data to this endpoint, we'll just reject it and lose the partial transaction information.

lahsivjar: This hasn't been implemented yet by any agent, in fact python agent is the first one to start implementing it so backward compatibility is not an issue at this point.

marclop (approved):
Code LGTM

lahsivjar:
@trentm @basepi Did you get a chance to test this out? I have done some testing with go-agent and it seems to work as expected but wanted to make sure all requirements are appropriately covered before merging.

trentm:
@lahsivjar I have not yet, but I think I may start looking at this tomorrow (as part of https://github.com/elastic/apm-agent-nodejs/issues/3136).

basepi:
I'll test this today.

basepi (approved):
Tested, working great!

PR #390, opened 2023-05-03
Prepare for v1.4.0 release

lahsivjar:
Release lambda extension `v1.4.0` with updated register transaction method.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -17,10 +17,10 @@
 ////
 
 [float]
-[[lambda-unreleased]]
-=== Unreleased
+[[lambda-1.4.0]]
+=== 1.4.0 - 2023/05/03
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.asciidoc
```diff
@@ -17,10 +17,10 @@
 ////
 
 [float]
-[[lambda-unreleased]]
-=== Unreleased
+[[lambda-1.4.0]]
+=== 1.4.0 - 2023/05/03
```
axw, line 21: Should we leave unreleased and add 1.4.0?

lahsivjar: That would add the unreleased section in the tag, I was planning to add it after releasing 1.4.0

axw: :+1: got it

axw (approved):

Issue #389, opened 2023-05-02
APM Server health check requests fail when using AWS secret manager


JonasKunz:
**APM AWS Lambda extensiom version** (`1.3.1`):

**Describe the bug**

When using the the [AWS Secret Manager](https://www.elastic.co/guide/en/apm/lambda/current/aws-lambda-secrets-manager.html) for providing the elastic APM secret token, the APM server health/version requests (`Get http://apm-server/`) are corrupted by the lambda extension.

It seems like an empty body is returned to the APM-agent with a 2xx status code:

```
2023-05-02 10:16:36,200 [elastic-apm-server-healthcheck] INFO  co.elastic.apm.agent.report.ApmServerHealthChecker - Elastic APM server is available: 
2023-05-02 10:16:36,204 [elastic-apm-server-healthcheck] WARN  co.elastic.apm.agent.report.ApmServerHealthChecker - Failed to parse version of APM server http://127.0.0.1:8200/: Unexpected end of JSON input
```

The java agent used here prints the response body of the health request after the `Elastic APM server is available: ` text, which is empty as shown here. 

This issue does not occur when directly passing in the secret token via environment variables instead.

PR #392, opened 2023-05-03
fix: send auth headers with info requests if missing

kruskall:
Forward auth headers for info requests. This does not affect data ingestion as we are already forwarding the headers there.

A test has been added to ensure the auth options passed to the lambda extension are correctly forwarded.

Closes https://github.com/elastic/apm-aws-lambda/issues/389

### DIFF
--- a/apmproxy/receiver.go
+++ b/apmproxy/receiver.go
@@ -107,6 +107,15 @@ func (c *Client) handleInfoRequest() (func(w http.ResponseWriter, r *http.Reques
 		r.Header.Set("X-Forwarded-Host", r.Header.Get("Host"))
 		r.Host = parsedApmServerURL.Host
 
+		// Send authorization header if the APM agent did not already forwarded it
+		if r.Header.Get("Authorization") == "" {
+			if c.ServerAPIKey != "" {
+				r.Header.Add("Authorization", "ApiKey "+c.ServerAPIKey)
+			} else if c.ServerSecretToken != "" {
+				r.Header.Add("Authorization", "Bearer "+c.ServerSecretToken)
+			}
+		}
+
 		// Forward request to the APM server
 		reverseProxy.ServeHTTP(w, r)
 	}, nil
--- a/apmproxy/receiver.go
+++ b/apmproxy/receiver.go
@@ -108,12 +108,10 @@ func (c *Client) handleInfoRequest() (func(w http.ResponseWriter, r *http.Reques
 		r.Host = parsedApmServerURL.Host
 
 		// Send authorization header if the APM agent did not already forwarded it
-		if r.Header.Get("Authorization") == "" {
-			if c.ServerAPIKey != "" {
-				r.Header.Add("Authorization", "ApiKey "+c.ServerAPIKey)
-			} else if c.ServerSecretToken != "" {
-				r.Header.Add("Authorization", "Bearer "+c.ServerSecretToken)
-			}
+		if c.ServerAPIKey != "" {
+			r.Header.Add("Authorization", "ApiKey "+c.ServerAPIKey)
+		} else if c.ServerSecretToken != "" {
+			r.Header.Add("Authorization", "Bearer "+c.ServerSecretToken)
 		}
 
 		// Forward request to the APM server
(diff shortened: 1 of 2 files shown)


apmproxy/receiver.go
```diff
@@ -107,6 +107,15 @@ func (c *Client) handleInfoRequest() (func(w http.ResponseWriter, r *http.Reques
 		r.Header.Set("X-Forwarded-Host", r.Header.Get("Host"))
 		r.Host = parsedApmServerURL.Host
 
+		// Send authorization header if the APM agent did not already forwarded it
+		if r.Header.Get("Authorization") == "" {
```
lahsivjar, line 111: Instead of checking if the `Authorization` header is set or not can we check if the secret manager is configured or not and override the headers if it is? In my mind, this seems to be a better approach than adding Auth headers if not provided but I might be mistaken entirely.

kruskall: I'm not sure about that approach.

This was adapted from what we are doing in the forwarder:

https://github.com/elastic/apm-aws-lambda/blob/bf8fc3fafb7ccb87f9a0e0c9cf97991a248f4010/apmproxy/apmserver.go#L163-L167

The client has not knowledge of aws  or the secrets manager. It only knows about the apikey and the secret token.

WDYT ?

lahsivjar: > This was adapted from what we are doing in the forwarder:

IIUC, the forwarder is different than the handle info request. The info request is a pure reverse proxy whereas the forwarder creates a new request and thus always adds the auth headers. I am a bit concerned about `if r.Header.Get("Authorization") == ""` condition; what do you think about overriding the auth headers unconditionally?

kruskall: good point! I updated the code!

apmproxy/receiver.go
```diff
@@ -107,6 +107,13 @@ func (c *Client) handleInfoRequest() (func(w http.ResponseWriter, r *http.Reques
 		r.Header.Set("X-Forwarded-Host", r.Header.Get("Host"))
 		r.Host = parsedApmServerURL.Host
 
+		// Send authorization header if the APM agent did not already forwarded it
```
lahsivjar, line 110:
```suggestion
		// Override authorization header sent by the APM agents
```

lahsivjar (approved):

lahsivjar:
Tested with a custom go lambda function which makes a call to `http://127.0.0.1:8200/` and logs the body. The function was configured to use AWS secrets manager ([ref](https://www.elastic.co/guide/en/apm/lambda/current/aws-lambda-secrets-manager.html)) instead of the secret token directly.

<details>

<summary>Used go lambda function</summary>

```go
package main

import (
	"context"
	"fmt"
	"io"
	"net/http"
	"strconv"
	"strings"

	"github.com/aws/aws-lambda-go/events"
	"github.com/aws/aws-lambda-go/lambda"

	_ "go.elastic.co/apm/module/apmlambda/v2"
)

var coldstart = true

func Handle(ctx context.Context, req events.APIGatewayProxyRequest) (events.APIGatewayProxyResponse, error) {
	var body strings.Builder
	resp, err := http.Get("http://127.0.0.1:8200/")
	if err != nil {
		body.WriteString(fmt.Sprintf("error in info request: %+v", err))
	} else {
		sb, _ := io.ReadAll(resp.Body)
		body.WriteString("body: ")
		body.Write(sb)
		body.WriteString(", ")
		body.WriteString("status_code: ")
		body.WriteString(strconv.Itoa(resp.StatusCode))
		body.WriteString(", ")
		body.WriteString("status: ")
		body.WriteString(resp.Status)
	}
	defer resp.Body.Close()
	response := events.APIGatewayProxyResponse{
		StatusCode: 200,
		Body:       body.String(),
		Headers: map[string]string{
			"coldstart": strconv.FormatBool(coldstart),
		},
	}
	coldstart = false
	return response, nil
}

func main() {
	lambda.Start(Handle)
}
```

</details>

<details>

<summary>Observed response without this PR</summary>

```
body: , status_code: 200, status: 200 OK
```
</details>

<details>

<summary>Observed response with this PR</summary>

```
body: {"build_date":"2023-02-13T13:01:54Z","build_sha":"8638b035d700e5e85e376252402b5375e4d4190b","publish_ready":true,"version":"8.6.2"}
, status_code: 200, status: 200 OK```
</details>

Issue #396, opened 2023-06-27
Use `UserAgent` with Lambda extension version


marclop:
**Is your feature request related to a problem? Please describe.**

Currently, when looking at APM Server logs it is hard to determine whether or not a Lambda Extension may be sending data to APM Server. The main reason is that the HTTP client is using the default `User-Agent` header (`Go-http-client/2.0`).

**Describe the solution you'd like**

It would be much better if we set the request's `User-Agent: APM-AWS-Lambda/<Version>`.

PR #404, opened 2023-09-06
feat: add user-agent with lambda extension version and propagate apm agent info

kruskall:
Add `User-Agent` header with lambda version and propagate apm agent info in the user agent when forwarding requests to apm-server

Closes https://github.com/elastic/apm-aws-lambda/issues/396


marclop (approved):
Thank you!

PR #414, opened 2023-09-29
Fix benchmark tf module with python function

lahsivjar:
- Adds support for using lambda layer in the terraform module.
- Fix extension zip path resolution.
- Update python lambda instrumentation to use lambda layer.


apmproxy/apmserver.go
```diff
@@ -96,10 +96,10 @@ func (c *Client) FlushAPMData(ctx context.Context) {
 	}
 
 	// Flush lambda data
+	c.logger.Debug("Flush in progress - Processing lambda data")
```
lahsivjar, line 99: [**For reviewers**] This log line led to a lot of noise during flush. Moved it so that only single entry is logged per flush.

carsonip (approved):

PR #423, opened 2023-12-22
 feat: add model package and drop go agent 

kruskall:
Define models for flexibility and avoid reusing go agent modules. 

fastjson is used to generate marshal methods


dmathieu (approved):
Nice!

PR #427, opened 2023-12-22
perf: load aws cfg lazily

kruskall:
Speed up init by lazily loading the AWS config. 

The config is only used when the secrets manager or an acm certificate is configured.


v1v:
run docs-build

PR #428, opened 2023-12-28
fix: upload two artifacts

kuisathaverat:
Recent changes in the upload artifacts action do not allow to update artifacts anymore so we have to upload two files with the tests results and process them separately.

see https://github.com/actions/toolkit/tree/main/packages/artifact#breaking-changes


amannocci (approved):
LGTM

PR #431, opened 2024-01-11
build: bump docker image deps

trentm:
- alpine 3.16.3 -> 3.16.8
- ensure that the release workflow gets the latest Go version, else it might get a not-quite-latest in a tool cache

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,6 +24,7 @@ jobs:
       - uses: actions/setup-go@v5
         with:
           go-version-file: 'go.mod'
+          check-latest: true
       - uses: docker/setup-qemu-action@v3
         with:
           platforms: linux/arm64, linux/amd64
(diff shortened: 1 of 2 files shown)


trentm:
Alpine 3.16.8 currently has no vulns: https://hub.docker.com/layers/library/alpine/3.16.8/images/sha256-ca452b8ab373e6de9c39d030870a52b8f0d3a9cf998c43183fd114660ae96330?context=explore
Another option would be to update to the latest Alpine 3.19, but I opted not to, because I think that we'd possibly want to do some more manual testing if doing that. I'm sure it would be fine, but still.

I added the "check-latest: true" to the release workflow because I noticed that the run for the last release (https://github.com/elastic/apm-aws-lambda/actions/runs/6542530791/job/17765697669) picked `go1.21.1` even though it ran on Oct 16, 2023 -- after go1.21.3 had been released (https://go.dev/doc/devel/release#go1.21.minor).

PR #434, opened 2024-01-22
feat: add `ELASTIC_APM_LAMBDA_DISABLE_LOGS_API` env var to disable logs api

kruskall:
Add flag to disable Logs API dubscription: `ELASTIC_APM_LAMBDA_DISABLE_LOGS_API`

Mostly for debugging. It should be tested before being documented as it's currently logging a warning on each event.

### DIFF
--- a/main.go
+++ b/main.go
@@ -47,7 +47,8 @@ func main() {
 		app.WithAWSConfig(cfg),
 	}
 
-	if os.Getenv("ELASTIC_APM_LAMBDA_DISABLE_LOGS_API") == "true" {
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -47,6 +47,10 @@ func main() {
 		app.WithAWSConfig(cfg),
 	}
 
+	if os.Getenv("ELASTIC_APM_LAMBDA_DISABLE_LOGS_API") == "true" {
```
carsonip, line 50: Should we use `strconv.ParseBool(...)` to be consistent with ELASTIC_APM_LAMBDA_CAPTURE_LOGS?

carsonip (approved):
thanks!

PR #437, opened 2024-01-23
test: use name_prefix to create unique aws resources

kruskall:
Use `name_prefix` instead of `name` to create unique aws resources and avoid smoke tests failing because resources already exist.

Upon investigation it looks like some resources were still around in AWS. I've deleted them manually and run the smoke test manually to confirm it works.

To reproduce: `EC_API_KEY=apikey TEST_DIR=tf make smoketest/run`

### DIFF
--- a/tf/.gitignore
+++ b/tf/.gitignore
@@ -1,5 +1,6 @@
 .terraform/
 
+.terraform.lock.hcl
 *.tfstate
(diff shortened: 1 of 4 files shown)


lahsivjar (approved):
I think we should migrate smoke tests to the `testing/tf-modules` eventually.

PR #459, opened 2024-04-22
github-action: use composite action and fix smoke-tests

v1v:
Fixes a regression when enabled the `goreleaser` with sbom. The smoke-tests uses `goreleaser` and I didn't detect it in https://github.com/elastic/apm-aws-lambda/blob/62236d32ae7613a4f82fefff471b09e4b1e1aca0/Makefile#L72

I refactored the code to use a composite action to reuse the same steps in the different GitHub workflows

### DIFF
--- a/.github/actions/bootstrap/action.yml
+++ b/.github/actions/bootstrap/action.yml
@@ -1,0 +1,31 @@
+---
+name: Bootstrap Checkout
+description: Ensures all actions bootstrap the same
+
+inputs:
+  goreleaser:
+    description: 'Install goreleaser toolchain ("true" or "false")'
+    required: false
+    default: "false"
+
+runs:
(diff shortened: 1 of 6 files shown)


.github/actions/bootstrap/action.yml
```diff
@@ -0,0 +1,31 @@
+---
```
v1v, line 1: The new GitHub composite action to be called

.github/dependabot.yml
```diff
@@ -28,3 +28,17 @@ updates:
       github-actions:
         patterns:
           - "*"
+
+  # GitHub composite actions
+  - package-ecosystem: "github-actions"
```
v1v, line 33: Enable dependabot for the composite actions

Issue #462, opened 2024-04-25
Lambda Layer appears to have been made unavailable


alexander-chi:
**APM AWS Lambda extensiom version** (`apm-aws-lambda version`):
arn:aws:lambda:eu-west-2:267093732750:layer:elastic-apm-extension-ver-1-5-3-x86_64:1

**Describe the bug**
When deploying functions using the above layer, we now receive the following error:

Resource handler returned message: "Layer version arn:aws:lambda:eu-west-2:267093732750:layer:elastic-apm-extension-ver-1-5-3-x86_64:1 does not exist. (Service: Lambda, Status Code: 400

**To Reproduce**
1. Deploy a function using the above lambda layer using CDK

**Expected behavior**
The layer in the release notes should exists

**Provide logs (if relevant)**:

Gawne:
We've noticed this and it's causing issues with production workloads. It's still in Elastic docs - https://www.elastic.co/guide/en/apm/agent/python/current/lambda-support.html

andrewbwhittle:
Same here.

alexander-chi:
The quick fix for us has been to just drop the layers all together and revert to calling APM start

alexander-chi:
It also appears that the 1.5.2 layers are still available, so reverting to those also works

dmathieu:
Thank you for reporting this.
It appears we had some workflow issues, where the release process ran again on newer commits, which deleted the layer and broke your versions.

Unfortunately, due to how lambda's releases work, we cannot fix the 1.5.3 layer, as any new layer would be tagged version 2.
We are contacting AWS's support to try undeleting the layer, but there's no guarantee they can do that.

So we have decided to release [1.5.4](https://github.com/elastic/apm-aws-lambda/releases/tag/v1.5.4) on 1980cea1de80bce1b1c93bd1d1eaf8eca4f9b13d, the same commit as 1.5.3.
So upgrading should be a noop and allow a fix of the layers.

Please let us know if that does fix your issue.

Gawne:
Thanks for the update @dmathieu. We'll upgrade and let you know how it goes

PR #463, opened 2024-04-26
Allow running release job with workflow_dispatch

dmathieu:
So we can rerun the release workflow against the proper tag, and fix the issue in #462.


dmathieu:
It seems opening PRs against forks isn't supported on this repository.
I don't have the permission to push a branch directly to the repo though.

PR #466, opened 2024-04-26
github-action: avoid release if github release already exists

v1v:
### What

If a github release exists, then it fails. Avoid rerunning the workflow since AWS lambdas are immutable and their versions are bumped if created a new one.

### Why
This should avoid the [edge case](https://github.com/elastic/apm-aws-lambda/issues/462) that a GitHub workflow was triggered accidentally and remove the lambdas.

### Test

image

### Further details

> github.ref_name	string	The short ref name of the branch or tag that triggered the workflow run. This value matches the branch or tag name shown on GitHub. For example, feature-branch-1.
For pull requests, the format is <pr_number>/merge.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,6 +22,14 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
+
+      - name: Validate tag does not exist on current commit
+        run: |
+          if [ $(git tag -l "${{ github.ref_name }}") ]; then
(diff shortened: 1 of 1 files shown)


dmathieu (dismissed):

reakaleek (dismissed):

dmathieu:
Won't the tag always already exist, since this action is triggered by its creation?

dmathieu (approved):

PR #510, opened 2024-07-25
fix: only flush logs if logs collection is enabled

kruskall:
If logs collection is disabled or logs subscription failed then `logsClient` is nil

### DIFF
--- a/app/run.go
+++ b/app/run.go
@@ -219,7 +219,6 @@ func (app *App) processEvent(
 		}()
 	} else {
 		app.logger.Warn("Logs collection not started due to earlier subscription failure")
-		close(logProcessingDone)
 	}
 
 	// Calculate how long to wait for a runtimeDoneSignal or AgentDoneSignal signal


rockdaboot:
LGTM, but can you also add a test for when there is no log collection?

app/run.go
```diff
@@ -96,8 +96,10 @@ func (app *App) Run(ctx context.Context) error {
 			backgroundDataSendWg.Wait()
 			if event.EventType == extension.Shutdown {
 				app.logger.Infof("Exiting due to shutdown event with reason %s", event.ShutdownReason)
-				// Flush buffered logs if any
-				app.logsClient.FlushData(ctx, event.RequestID, event.InvokedFunctionArn, app.apmClient.LambdaDataChannel, true)
+				if app.logsClient != nil {
```
rockdaboot, line 99: Alternatively, check `lc` for nil in `FlushData` and return early (slightly reduces number of lines changed).

rockdaboot (approved):

PR #511, opened 2024-07-25
test: create a new bytes reader instead of sharing bytes buffer

kruskall:
http requests are sharing a bytes buffer causing the request to fail with

Post "http://localhost:39859/intake/v2/events": http: ContentLength=1082 with Body length 0

Update test to create a bytes reader for each request and check the error to avoid failing silently.
Close the response body to avoid leaks or leaving the client in a bad state.


rockdaboot (approved):
LGTM

PR #524, opened 2024-07-31
test: Use automatically created GITHUB_TOKEN for smoketest

rockdaboot:
This hopefully fixes the error from `gh` in the smoketest:
```
Program: ./get_username.sh
Error Message: gh: Resource not accessible by integration (HTTP 403)
```

### DIFF
--- a/.github/workflows/smoke-tests.yml
+++ b/.github/workflows/smoke-tests.yml
@@ -20,7 +20,6 @@ jobs:
   test:
     runs-on: ubuntu-latest
     env:
-      GH_TOKEN: ${{ github.token }}
       SSH_KEY: "./id_rsa_terraform"
       TF_VAR_private_key: "./id_rsa_terraform"
       TF_VAR_public_key: "./id_rsa_terraform.pub"
--- a/testing/smoketest/get_username.sh
+++ b/testing/smoketest/get_username.sh
@@ -1,7 +1,11 @@
 #!/bin/bash -eu
 
+: ${CI:=""}
+
+if [ -z "$CI" ]; then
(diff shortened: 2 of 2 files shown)


.github/workflows/smoke-tests.yml
```diff
@@ -20,7 +20,7 @@ jobs:
   test:
     runs-on: ubuntu-latest
     env:
-      GH_TOKEN: ${{ github.token }}
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
reakaleek, line 23: I'm afraid this is just another syntax to get the same token.

We would need to grant the token the correct [permissions](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/assigning-permissions-to-jobs).

However, unfortunately, I don't see appropriate permission to query the GH user API.

reakaleek: As an alternative, we could use a machine user token.

I don't know what the username is used for, but could we do something like:

```
if [[ -n "$CI"]]; then 
  user_name = "github-actions"
else 
# ...
fi
```
in https://github.com/elastic/apm-aws-lambda/blob/54ff53aaaefba7d67d93a71bbca14631749df3d5/testing/smoketest/get_username.sh#L11-L10 ?

rockdaboot: That's a good suggestion. Using a hard-coded user_name in CI should be fine.

reakaleek (approved):

rockdaboot:
@reakaleek Sorry, I messed this up. Fix at #525

PR #570, opened 2024-10-25
Avoid race conditions when handling data

florianl:
Initialize local variable only once and make sure expected APMData is handled first.

### DIFF
--- a/accumulator/batch.go
+++ b/accumulator/batch.go
@@ -37,6 +37,8 @@ var (
 	// ErrInvalidEncoding is returned for any APMData that is encoded
 	// with any encoding format
 	ErrInvalidEncoding = errors.New("encoded data not supported")
+	// ErrNoData indicates that APMData.data is empty
+	ErrNoData = errors.New("no data")
 )
 
 var (
@@ -170,7 +172,7 @@ func (b *Batch) OnAgentInit(reqID, contentEncoding string, raw []byte) error {
 // before adding any events then ErrBatchFull is returned.
 func (b *Batch) AddAgentData(apmData APMData) error {
 	if len(apmData.Data) == 0 {
-		return nil
+		return ErrNoData
 	}
 	raw, err := GetUncompressedBytes(apmData.Data, apmData.ContentEncoding)
 	if err != nil {
--- a/apmproxy/apmserver.go
+++ b/apmproxy/apmserver.go
@@ -60,6 +60,10 @@ func (c *Client) ForwardApmData(ctx context.Context) error {
 			return nil
 		case data := <-c.AgentDataChannel:
 			if err := c.forwardAgentData(ctx, data); err != nil {
+				if errors.Is(err, accumulator.ErrNoData) {
+					c.logger.Debug("received something from AgentDataChannel without APMData")
+					continue
+				}
 				return err
 			}
 			// Wait for metadata to be available, metadata will be available as soon as
--- a/apmproxy/apmserver.go
+++ b/apmproxy/apmserver.go
@@ -61,11 +61,11 @@ func (c *Client) ForwardApmData(ctx context.Context) error {
 			c.logger.Debug("Invocation context cancelled, not processing any more agent data")
 			return nil
 		case data := <-c.AgentDataChannel:
+			if len(data.Data) == 0 {
+				c.logger.Debugf("Received something from '%s' without APMData", data.AgentInfo)
+				continue
+			}
 			if err := c.forwardAgentData(ctx, data); err != nil {
-				if errors.Is(err, accumulator.ErrNoData) {
-					c.logger.Debugf("Received something from '%s' without APMData", data.AgentInfo)
-					continue
-				}
 				return err
 			}
 			once.Do(func() {
--- a/apmproxy/apmserver.go
+++ b/apmproxy/apmserver.go
@@ -63,7 +63,7 @@ func (c *Client) ForwardApmData(ctx context.Context) error {
 		case data := <-c.AgentDataChannel:
 			if err := c.forwardAgentData(ctx, data); err != nil {
 				if errors.Is(err, accumulator.ErrNoData) {
-					c.logger.Debug("received something from AgentDataChannel without APMData")
+					c.logger.Debugf("received something from '%s' without APMData", data.AgentInfo)
 					continue
 				}
 				return err
--- a/apmproxy/apmserver.go
+++ b/apmproxy/apmserver.go
@@ -29,6 +29,7 @@ import (
 	"math/rand"
 	"net/http"
 	"time"
+	"sync"
 
 	"github.com/elastic/apm-aws-lambda/accumulator"
 	"github.com/elastic/apm-aws-lambda/version"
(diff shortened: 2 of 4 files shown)


apmproxy/apmserver.go
```diff
@@ -60,11 +62,17 @@ func (c *Client) ForwardApmData(ctx context.Context) error {
...
+				if errors.Is(err, accumulator.ErrNoData) {
+					c.logger.Debugf("received something from '%s' without APMData", data.AgentInfo)
+					continue
+				}
 				return err
 			}
-			// Wait for metadata to be available, metadata will be available as soon as
-			// the first agent data is processed.
-			lambdaDataChan = c.LambdaDataChannel
+			once.Do(func() {
```
christos68k, line 71: Not sure what's happening here, why don't we use `c.LambdaDataChannel` directly and have to first assign it to this local variable? Isn't this a NOP ?

florianl: @kruskall or @vishal-jogi maybe can provide more insights on this design decission in the first place.
With `sync.once` the local variable should be initialized exactly once instead of every time with some received from `c.AgentDataChannel`. I just have learned, that `c.LambdaDataChannel` should not change over time.

apmproxy/apmserver.go
```diff
@@ -60,11 +62,17 @@ func (c *Client) ForwardApmData(ctx context.Context) error {
...
+				// Wait for metadata to be available, metadata will be available as soon as
+				// the first agent data is processed.
+				lambdaDataChan = c.LambdaDataChannel
+			})
 		case data := <-lambdaDataChan:
```
christos68k, line 76: If `c.LambdaDataChannel` doesn't change then why don't we use it directly? Unless there are hidden semantics, both the channel assignment and the `Once.Do` are essentially NOPs. The comment is confusing me, as there's no waiting we're just assigning a reference.

```suggestion
		case data := <-c.lambdaDataChan:
```

apmproxy/apmserver.go
lahsivjar: I don't see why we require a sync primitive here. There is no race: `lambdaDataChan` is a local variable and `c.AgentDataChannel` is created only once. If we really want to do the initialization of the local variable only once then we can just do a nil check or something.

> why don't we use c.LambdaDataChannel directly and have to first assign it to this local variable? Isn't this a NOP ?

We don't use the `c.LambdaDataChannel` because we don't want to consume from the lambda API until we are sure we can process the data and send it to APM server -- this means that we have the metadata available from the APM agents. This allows us to pushback to the lambda API rather than consuming data without metadata being available and losing the data.

christos68k: Ok makes sense.

apmproxy/apmserver.go
```diff
@@ -60,11 +62,17 @@ func (c *Client) ForwardApmData(ctx context.Context) error {
 			return nil
 		case data := <-c.AgentDataChannel:
 			if err := c.forwardAgentData(ctx, data); err != nil {
+				if errors.Is(err, accumulator.ErrNoData) {
+					c.logger.Debugf("Received something from '%s' without APMData", data.AgentInfo)
+					continue
+				}
```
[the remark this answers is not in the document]
florianl: good point, will update

apmproxy/apmserver.go
lahsivjar: Agreed, I would recommend `nil` check because there is no need for a `sync` here

florianl: I have updated the comment. Hope it clarifies the assumptions and expectations at this point.

rockdaboot: Hm, I still don't understand why a `nil` check is not sufficient here.

accumulator/batch.go
```diff
@@ -37,6 +37,8 @@ var (
 	// ErrInvalidEncoding is returned for any APMData that is encoded
 	// with any encoding format
 	ErrInvalidEncoding = errors.New("encoded data not supported")
+	// ErrNoData indicates that APMData.data is empty
+	ErrNoData = errors.New("no data")
```
lahsivjar, line 41: I am not sure about this, no data is not an error for the batch and I don't like returning no data errors. Now that we have already handled no data case in `ForwardAPMData`, can we remove this?

florianl: What is your suggestion to return for the`len(apmData.Data) == 0` case in `AddAgentData()`?
Just because there is a check in `ForwardAPMData()` does not prevent other users of `AddAgentData()` running into this issue.

lahsivjar: My point is that, for the batch having an empty data is not an error. The batch can just not add data in this case. If we want to protect against edge cases we should do it at the source (like how we do in `ForwardAPMData`).

lahsivjar: Also, if we want to be more specific in handling the behaviour, we could introduce another method in the batch to check the metadata status, something like `bool HasMetadata()` and incorporate it where we require such a check.

florianl: I still consider the use case `len(apmData.Data) == 0` in `AddAgentData()` dangerous even if the only caller atm :tm: is using a check. What makes this package and its API safe to use, if something else calls `AddAgentData()`?
Even if no data is considered a non error case, what should it be considered instead and how should it be treated? With the current behavior it seems to run into issues.

>  we could introduce another method in the batch to check the metadata status, something like bool HasMetadata() and incorporate it where we require such a check.

If `AddAgentData()` implements such a check, what should be returned in a case where
- there is not metadata but `len(apmData.Data) != 0`. Is in this case data loss the expected result?
- there is metadata but `len(apmData.Data) == 0`

lahsivjar: > What makes this package and its API safe to use, if something else calls AddAgentData()?

Agent data here refers to data from the APM agents and intake-v2 is the only supported protocol. What other cases do you have in mind here?

> there is not metadata but len(apmData.Data) != 0. Is in this case data loss the expected result?

If data from the agent is not nil then it MUST have the metadata as the first line.

> there is metadata but len(apmData.Data) == 0

`apmData.Data` is including the metadata so this is also not possible. We could have only metadata in the body and that is okay.

christos68k: @lahsivjar So essentially in your view, `ErrNoData` and associated logic is redundant as "no data" can take place during normal operation, and if we want to check for this case, we should do it before calling `AddAgentData`. Is that correct?

christos68k (commented):
I think the fix that @florianl implements in this PR (not exiting the processing loop when `len(data.Data) == 0` is important and should be merged. @lahsivjar Do we need anything else to get this merged?

accumulator/batch.go
lahsivjar: Basically, yes! I don't have a strong opinion here so feel free to ignore it.

lahsivjar:
> Do we need anything else to get this merged?

I think `sync.Once` to `nil` check we should be good to merge. The remaining point on the `ErrNoData` is not that important and we can ignore it if it doesn't make sense to others.

apmproxy/apmserver.go
florianl: Updated the PR to use a `nil` check instead of `sync.Once()` with https://github.com/elastic/apm-aws-lambda/pull/570/commits/1763b69bc212ca52f5c8942eb2359b2e15cbbfec.

christos68k (approved):

PR #613, opened 2025-02-18
fix: forward logs directly when invocation is over

kruskall:
The lambda extension process data in the background during an invocation:

https://github.com/elastic/apm-aws-lambda/blob/3f5620c0e1e824c6db6d2172cb29a127bc21cd1f/app/run.go#L197-L204

Data is sent to a channel. This is fine during an invocation when the goroutine is running but it's possible that during shutdown the channel will block because there's no goroutine running.

https://github.com/elastic/apm-aws-lambda/blob/3f5620c0e1e824c6db6d2172cb29a127bc21cd1f/app/run.go#L101

The solution is to bypass the channel and forward the data directly during shutdown.


logsapi/event.go
```diff
@@ -60,13 +62,13 @@ func (lc *Client) ProcessLogs(
 	ctx context.Context,
 	requestID string,
 	invokedFnArn string,
-	dataChan chan []byte,
+	forwarFn Forwarder,
```
rockdaboot, line 65:
```suggestion
	forwardFn Forwarder,
```

logsapi/event.go
```diff
@@ -60,13 +62,13 @@ func (lc *Client) ProcessLogs(
...
 	isShutdown bool,
 ) {
 	for {
 		select {
 		case logEvent := <-lc.logsChannel:
-			if shouldExit := lc.handleEvent(ctx, logEvent, requestID, invokedFnArn, dataChan, isShutdown); shouldExit {
+			if shouldExit := lc.handleEvent(ctx, logEvent, requestID, invokedFnArn, forwarFn, isShutdown); shouldExit {
```
rockdaboot, line 71:
```suggestion
			if shouldExit := lc.handleEvent(ctx, logEvent, requestID, invokedFnArn, forwardFn, isShutdown); shouldExit {
```

dmathieu (commented):
Could we add tests, or is that too unpredictable?

rockdaboot:
> Updated! 👍

Thanks! Maybe I misunderstand the code changes, but it looks to me that you replace direct writes to the channel with a function that writes to the data channel. How is that a difference in the data flow logic and don't we still have the same issue then (channel reader doesn't read any more)?

I somewhat agree with @dmathieu that we should try to create a test that reproduces the issue, and then prove that this code solves the issue. It doesn't have to be a unit test, if that is too involved. Maybe some kind of tool / setup / script, whatever is the simplest way.

rockdaboot:
> For the third case: It's fine to send to a channel inside processEvent because there's someone reading from it and in the worst case the ctx will make the func exit.

Thanks for the explanation.

> maybe we can reuse the setup from @xrmx ? I'm not sure how feasible it is to write a unit test for this.

@xrmx Can you run your reproducer again with the changes here and confirm that it fixes the issue?

dmathieu (approved):

PR #624, opened 2025-03-13
Update axios to 1.8.2 in e2e tests

rockdaboot:
Fixes https://github.com/elastic/apm-aws-lambda/security/dependabot/43

### DIFF
--- a/e2e-testing/sam-nodejs/sam-testing-nodejs/package.json
+++ b/e2e-testing/sam-nodejs/sam-testing-nodejs/package.json
@@ -7,7 +7,7 @@
   "author": "SAM CLI",
   "license": "MIT",
   "dependencies": {
-    "axios": "^0.28.1",
+    "axios": "^1.8.2",
     "elastic-apm-node": "^3.26.0"
   },
   "scripts": {


dmathieu:
This goes a major bump. Are we sure there aren't any untested breaking changes?
cc @trentm

trentm:
I guess I don't have perms on this repo to view https://github.com/elastic/apm-aws-lambda/security/dependabot/43

trentm:
These e2e-testing bits are, I gather, very out of date. It looks like the e2e-testing bits aren't run regularly.

- this axios update is fine, the small "app.js" in this package isn't even *using* the `axios` dep, so it could be removed.

This `sam-testing-nodejs` dir was adapted from a small aws-sam-cli test package that now lives here: https://github.com/aws/aws-sam-cli/tree/develop/samcli/lib/init/templates/cookiecutter-aws-sam-hello-nodejs/%7B%7Bcookiecutter.project_name%7D%7D/hello-world
That includes a dep on `axios`, but has any code using `axios` commented out.
The adapted copy here did not copy over the unit tests, so:

- the `devDependencies` could all be removed as well

- As well, the `elastic-apm-node` dependency is on an *old major*, the current supported versions are `4.x`.

@rockdaboot Do you have a preference here? We could (a) take this PR as is; (b) use this or a separate PR to make these other updates. I'm happy to do the quick updates for (b).

rockdaboot:
> Do you have a preference here? We could (a) take this PR as is; (b) use this or a separate PR to make these other updates. I'm happy to do the quick updates for (b).

@trentm It looks like (b) would be the proper thing to do. I#d would be great if you can do it!

trentm:
> It looks like (b) would be the proper thing to do. I'd would be great if you can do it!

https://github.com/elastic/apm-aws-lambda/pull/626

PR #642, opened 2025-04-09
feat: bump go to 1.24 and tool deps

kruskall:
use native go tool directive
bump tool dependencies
fix linter issues
run terraform-docs


rockdaboot:
Is this still relevant after merging #644?

PR #664, opened 2025-05-19
feat: update notice file generation and move licenser to go tool

kruskall:
move licenser to go tool to make use of the module system

update notice file template to comply with elastic policy/template

regenerate files


rockdaboot:
@kruskall Can you add your intention, maybe with pros/cons to the PR description? Otherwise, a reader has to imagine or "reverse-engineer" this. Thank you 🙏🏻

rockdaboot (approved):

